PR #6, opened 2019-02-12
Merge in algorithms from vg

ekg:
Here I'm threading algorithms into a command line interface. The goal is to get dsgvg into shape as a graph processor that we can use to work on really big, non-divisible graphs.

@adamnovak sorry about this. In the end, I have had to pull these in directly. I need to make a bunch of minor changes and it's not sensible for me to iterate through the vg PR pathway to do it at this point.

I've broken my handle graph interface slightly because I don't have a "swap" function. I'm contemplating adding this back in because in principle it could let us save a copy of the structure to apply the ordering, but I suspect it will be very slow in comparison.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,8 +4,8 @@ cmake_minimum_required(VERSION 3.1)
 
 # Project's name
 project(dankgraph)
-# We build using c++11
-set(CMAKE_CXX_STANDARD 11)
+# We build using c++14
+set(CMAKE_CXX_STANDARD 14)
 
 # Use all standard-compliant optimizations
 set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O3 -g")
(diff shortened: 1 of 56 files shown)


adamnovak:
Did you add the algorithms files unmodified in one commit and make the changes in another? That would probably help a lot for eventually unifying the copies.

adamnovak:
Also, @jeizenga wants to drop `swap` and the integrated ordering from `HandleGraph`, so it's fine to not implement it here I think.

jeizenga:
Fwiw, both `HashGraph` and `PackedGraph` are non-compliant with the documentation on `swap_handles`.

ekg:
> Did you add the algorithms files unmodified in one commit and make the changes in another? That would probably help a lot for eventually unifying the copies.

Sorry, I didn't :disappointed: 

The differences are relatively minimal, and probably we'll be able to see what's up with a quick diff.

PR #7, opened 2019-02-13
total rewrite as bytegraph

ekg:
Functions for enumerating kmers on the graph. This might result in some optimizations to the dsgvg index itself.

I'm picking up a lot of this from vg. It's basically a port of similar functions in vg but without support for GCSA2 style kmer doubling or kmer output. For efficiency, it might be best to pull that code back in separately.

Initially, my goal is kmer enumeration as input to sparse or sampled kmer indexes. For sparse indexes, I'm hoping to use modimizers (recording only the subset of kmers for which hash(kmer) % N == 0).


ekg:
I've rewritten the internals almost completely to store the graph relative to nodes, with each node represented as a packed set of bytes (`std::vector<uint8_t>`). This is much faster. I'm currently pulling in @adamnovak's libhandlegraph update from master.

Then, I need a better name for this tool. Maybe something that we can pronounce.

PR #20, opened 2019-05-15
Sort better

ekg:
Fixes some problems with sorting.

Tries a bunch of things that don't work to improve sort order.

I want to explore using paths as hints.

### DIFF
--- a/src/algorithms/eades_algorithm.cpp
+++ b/src/algorithms/eades_algorithm.cpp
@@ -15,11 +15,17 @@ using namespace handlegraph;
 #endif
         
         // decide which strand will be "forward" for each node
+        /*
         vector<handle_t> canonical_orientation = single_stranded_orientation(graph);
         if (canonical_orientation.size() < graph->node_size()) {
             cerr << "error:[eades_algorithm] Eades' algorithm only valid on graphs with a single stranded orientation" << endl;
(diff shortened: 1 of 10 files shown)


src/algorithms/eades_algorithm.cpp
```diff
@@ -15,11 +15,17 @@ using namespace handlegraph;
 #endif
         
         // decide which strand will be "forward" for each node
+        /*
         vector<handle_t> canonical_orientation = single_stranded_orientation(graph);
         if (canonical_orientation.size() < graph->node_size()) {
             cerr << "error:[eades_algorithm] Eades' algorithm only valid on graphs with a single stranded orientation" << endl;
             exit(1);
         }
+        */
```
jeizenga: The point of the canonical orientation is to allow graphs that are still single stranded to "reverse" in certain segments, as long as there are no reversing cycles. If you're going to replace it with the handle's forward orientation you could do away with the canonical orientation entirely and just check whether the handles are oriented forward. It's worth noting however, that this change does make the algorithm less general, so that the graph needs to be single stranded as well as have no reversing edges.

ekg:
Thanks for this, I was assuming what you are describing. I wanted to get a
sense of the kind of sort order this would give me.

PR #26, opened 2019-06-11
the mapper: it begins

ekg:
`odgi` is going to become a sequence to graph mapper. Here's how:

1. two pass kmer indexing via bbhash (to map kmers to a smaller index range) and dynamic (to store positions compactly)
2. implement a path position index / store in position table
4. clustering concepts from vg, with appropriate twists
5. dozeu based x-drop DP alignment to the graph, driven by A* algorithm
6. output GAF... (should confirm format with @lh3)
7. provide conversion from GAF to .gam and .bam ("surjection")

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,6 +220,26 @@ ExternalProject_Add(picosha256
 ExternalProject_Get_property(picosha256 SOURCE_DIR)
 set(picosha256_INCLUDE "${SOURCE_DIR}")
 
+# mmmultimap (memory mapped multimap)
+ExternalProject_Add(mmmultimap
+  GIT_REPOSITORY "https://github.com/ekg/mmmultimap.git"
+  GIT_TAG "9a7c6dc3550c5fc495a3331cc7c3219ae574b5ca"
+  BUILD_COMMAND ""
+  UPDATE_COMMAND ""
+  INSTALL_COMMAND "")
+ExternalProject_Get_property(mmmultimap SOURCE_DIR)
+set(mmmultimap_INCLUDE "${SOURCE_DIR}/src")
+
+# In-place Parallel Super Scalar Samplesort (IPS⁴o), header only
+ExternalProject_Add(ips4o
+  GIT_REPOSITORY "https://github.com/SaschaWitt/ips4o.git"
+  GIT_TAG "bff3ccf0bf349497f2bb10f825d160b792236367"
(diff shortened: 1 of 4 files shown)


edawson:
> output GAF... (should confirm format with @lh3)

Is there a definition out there for this format? There's also https://github.com/edawson/gram and https://docbase.io/posts/300364/sharing/108ce773-f90c-4d8e-8ae7-ad0456105fe6 being tossed around. I'd love to have another public spec a la GFA.

lh3:
See [here](http://www.liheng.org/downloads/rGFA-GAF.pdf) (PDF) for brief description.

ekg:
Thanks, I think I can work towards this.

I was hoping to use a simpler model that doesn't require the reference
paths to fully cover the graph, as this isn't always possible. Here the
reference contig name in PAF would be replaced by a series of node IDs or
names, and otherwise it would basically be the same semantics. I guess we
can fall back to this behavior in graphs without embedded paths.

I would call this GAF, and the version you describe rGAF. We can convert
between the two when we have embedded reference paths in the graph. But if
GAF is only depending on node IDs then we don't need to add paths to the
GFA (making rGFA) if we don't have them to add. This is a common issue for
many graphs we are building. I'm not sure of the best way to resolve this,
but I guess we can work things out with examples and see how it goes.

Another possibility is to extend the format to provide graph to graph
mappings. In this case both the query and target sequence would be
represented with the same semantics. What do you think about this? I don't
want to get too far ahead of working examples so that any discussion stays
grounded, but in any case I'm curious if you have any ideas. We have a
Translation concept from vg which is similar in scope and it's been
important in a lot of operations in variant calling. CC @glennhickey

lh3:
In GAF, a path like `>ctg:100-200` indicates a stable path; a path like `>v1<v2` (i.e. without start and end) indicates a path in the vertex coordinate. This has been planned all along. A caveat with this notation is that a vertex name can't end with `:number1-number2`; otherwise, it would be impossible to distinguish the two cases. The PDF doesn't explain this part to avoid potential complications.

I haven't thought much about graph-to-graph mappings.

ekg:
Ah, I forgot about this being part of the general plan. The restriction on
sequence naming is no problem.

We should start a repository to maintain the format definitions. There is
no rush, but it will become important when we have a few tools reading and
writing this data type. I'm not sure if GFA-spec is the right place.

PR #42, opened 2019-09-05
adjust binning to be more useful for interactive visualization

ekg:
Demo to come at BioHackathon 2019.


ekg:
To generate this binning in JSON:

```
odgi build -g DRB1-3123.gfa -o - | odgi sort -i - -o DRB1-3123.og
for i in 1 10 100 1000; do odgi bin -i DRB1-3123.og -w $i -j | pigz >DRB1-3123.og.bin$i.json.gz; done
```

PR #43, opened 2019-09-08
more improvements to sorting

ekg:
Thus far, this branch implements a new kind of sort which I'm calling "dagified sort". In this, we dagify the graph and use the sort positions we get there to order the nodes in the original graph. This deals with the somewhat pathological problem that our pseudo topological sorter often pushes cyclic components very far in the sort order from their actual topological position. There might be better solutions.

In this branch I'll also be pulling in a bunch of changes that let us sort the graph using sparse matrix diagonalization techniques. The idea is that these are used for neighborhood detection in graphs, and these should still be applicable to genome graphs despite the relative sparsity of the edge matrix. For instance, we hope that a chromosome or SV cluster would tend to form a neighborhood in such a representation, without being affected by cycles or inversions as much as the methods we have available are.

These methods are great for getting the large scale order of the graph into a more meaningful shape, but aren't so good when dealing with local structures. I've discovered that I get great performance out of a series of sorts (eades, DFS, DAGified, topo) after a sort based on a symmetric (row and column equivalent) permutation of the sparse matrix representing an undirected version of the graph. These are seeded with the order from the matrix permutation and that's then smoothed by the locally correct topology-respecting properties of these methods. Most of the graphs are in acyclic segments, and exact sorting of these does best.

So for this graph of DRB1-3123, which is slightly tangled when built with minimap2 and seqwish, I can get a sort that seems to respect the overall topology (as implied by Bandage's force directed layout):

image

image

To do this, I'm applying particular configurations of the [Mondriaan](https://git.science.uu.nl/R.H.Bisseling/mondriaan) package for sparse matrix partitioning. One tricky aspect of this is that the partitioning seems to require a fixed parameter for the number of partitions. There are also lots and lots of options about how this is done.

This approach appears to have very nice runtime behavior, and runs in about the same amount of time as the series of deterministic sort algorithms that I apply post-hoc to clean up the fine structure of the sorted order.

This might be of interest to @jeizenga, @adamnovak, and @haussler.

Here's a script that should roughly reproduce the test on DRB1-3123:

```
# build the graph
minimap2 -c -x asm20 -X -r 10 -t 4 DRB1-3123.fa DRB1-3123.fa | gzip >DRB1-3123.paf.gz && time seqwish -s DRB1-3123.fa -p DRB1-3123.paf.gz -b DRB1-3123.work -g DRB1-3123.gfa
# make the sparse matrix form of it
odgi build -g DRB1-3123.gfa >DRB1-3123.og
odgi matrix -i DRB1-3123.og > DRB1-3123.mtx
# use mondrian to derive a sort
Mondriaan DRB1-3123.mtx 8 1 -Permute=SBD -EnforceSymmetricPermutation=yes -Coarsening_StopRatio=1 -SplitStrategy=finegrain -Iterative_Refinement=always -Coarsening_InprodScaling=max -VectorPartition_Step3=decrease -Coarsening_InprodMatchingOrder=incrwgt -Coarsening_NetScaling=no -Coarsening_MatchingStrategy=ata -Coarsening_MatchingATAMatcher=greedy -SquareMatrix_DistributeVectorsEqual_AddDummies=no -ImproveFreeNonzeros=no -LoadbalanceStrategy=increase 
# apply the sort and clean it up
odgi sort -i DRB1-3123.og -s DRB1-3123.mtx-col8 -o - \
    | odgi sort -i - -o - -e \
    | odgi sort -i - -o - -b \
    | odgi sort -i - -o - -d \
    | odgi sort -i - -o - \
    | odgi viz -i - -o DRB1-3123.gfa.viz.png -x 2000 -R
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -222,6 +222,17 @@ ExternalProject_Add(picosha256
 ExternalProject_Get_property(picosha256 SOURCE_DIR)
 set(picosha256_INCLUDE "${SOURCE_DIR}")
 
+# SGD based graph layout
+ExternalProject_Add(sgd2
+  GIT_REPOSITORY "https://github.com/ekg/sgd2.git"
+  GIT_TAG "3cb6ffe65dc14bf580866519e62d9b52964bdcae"
+  UPDATE_COMMAND ""
+  INSTALL_COMMAND ""
+  BUILD_COMMAND ""
+  CONFIGURE_COMMAND "")
+ExternalProject_Get_property(sgd2 SOURCE_DIR)
+set(sgd2_INCLUDE "${SOURCE_DIR}/src")
+
 # pybind11
 #ExternalProject_Add(pybind11
 #    GIT_REPOSITORY https://github.com/pybind/pybind11.git 
@@ -304,7 +315,8 @@ set(odgi_DEPS
     lodepng
     structures
     sonlib
-    picosha256)
+    picosha256
+    sgd2)
 add_dependencies(odgi_objs ${odgi_DEPS})
 
 set(odgi_INCLUDES
@@ -326,7 +338,8 @@ set(odgi_INCLUDES
   "${structures_INCLUDE}"
   "${sonLib_lib_INCLUDE}"
   "${sonLib_inc_INCLUDE}"
-  "${picosha256_INCLUDE}")
+  "${picosha256_INCLUDE}"
+  "${sgd2_INCLUDE}")
 
 set(odgi_LIBS
   "${sdsl-lite_LIB}/libsdsl.a"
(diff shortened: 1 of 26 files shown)


ekg:
It's now possible to implement sorting pipelines in odgi, for instance:

```
odgi build -g DRB1-3123.gfa -o - \
    | odgi sort -i - -o - -p embdfdfn -N 500 -E 1 \
    | odgi viz -i - -o x.png -x 2000 -y 400 -R
```

Here, we first run Eades algorithm, then Mondriaan with 500 partitions and epsilon of 1, then the "cycle breaking" sort, then "dagsort" (DAGify the graph then take the mean sort position of the nodes in that graph to determine the sort order), flip (just reverse the graph), dagsort, flip, and "unseeded" topological sort.

The result is nice.

image

This is a solution to improving sort quality. Unfortunately, it opens the floodgates to many arbitrary sorting pipelines. What we want is something more principled, based on path positions and correspondence in the graph.

I still see that topological sorting puts inverting and cycling small components at weird places in the graph. Ultimately, we want a kind of optimization procedure that penalizes this. However, these are expensive.

PR #57, opened 2019-11-03
Morepybind1 > WIP too discus not to merge as is

JervenBolleman:
Hi @ekg,

I have been discovering a bit about how things are done in odgi. However, not everything is made available in the excellent python bindings. I started work on making some of this available. Not all sure though, if any of this makes sense.

Regards,
Jerven

### DIFF
--- a/src/pythonmodule.cpp
+++ b/src/pythonmodule.cpp
@@ -22,7 +22,7 @@ PYBIND11_MODULE(odgi, m)
         .def("path_id", [](handlegraph::step_handle_t &step_handle) {
                 return (reinterpret_cast<const int64_t*>(&step_handle)[0]) >> 1;
         })
-        .def("is_Reverse", [](handlegraph::step_handle_t &step_handle) {
+        .def("is_reverse", [](handlegraph::step_handle_t &step_handle) {
                 return (reinterpret_cast<const int64_t*>(&step_handle)[0]) & 1;
         })
         .def("prev_id", [](handlegraph::step_handle_t &step_handle) {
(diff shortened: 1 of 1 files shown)


ekg:
@JervenBolleman this is beautiful.

As a major user of this (and probably a much better python coder than me), I'll follow your lead in this. Please let me know when you think we should merge this.

Do you have any questions about other things we should add? You mentioned iterators by email. I'm not yet sure how to do that.

JervenBolleman:
@ekg, this is a more general question. Does it make sense to have python bindings for libhandlegraph which are shared between different implementations. Also how much am I breaking different parts of the api by surfacing parts in areas while there is a defined way to do it in the handlegraph api that odgi implements. 

For the rest my code is called python programming by stackoverflow ;)

ekg:
I think it makes sense to make python bindings that are shared across all of libhandlegraph implementations. But, I don't think it's possible to do this with a single implementation of the pybind11 interface. It seems we have to do one per implementation. The alternative is to somehow build a forwarding interface that abstracts over different possible backing implementations. Perhaps someone else understands this better?

But, if you keep to exposing functions and data structures that are inside the libhandlegraph interface, then we can easily copy/paste this code for each of the other implementations. We can even do this in some kind of scriptable way by replacing the class names appropriately. Does this make sense?

subwaystation:
Hi @JervenBolleman 
this has been open for a while. Do you still need it?
@ekg Are there any downstream project that rely on the python bindings except SpOdgi?
Else I would just merge it in.

JervenBolleman:
I think it can be merged in as is, I doubt it would hurt anyone. I also doubt that I will take the time to work further on this in the coming year.

PR #72, opened 2020-01-24
Documentation

ctcisar:
Added tutorial in readthedocs format in /docs


adamnovak:
Hey @ctcisar! Thanks for all of this, great work!

I have a couple of questions I want answered for maintainability, though.

First, do you have source files for the PNG figures at all? Like SVGs or the files from whatever program you made them in? Having those available would help if they need to be changed for any reason, and I'm thinking of changing your bidirectional edge notation a bit to make it look more like there are two edges and a visible circular loop.

Second, your glossary splitting out all the methods by what they operate on is an amazing idea, but did you really just type it all out by hand? Or did Binder or Sphinx do something to help pull in the method descriptions from C++ and combine the overloads and so on? I would want the method descriptions to update when we change them in the C++ source; otherwise I foresee things getting out of sync. If you did just type them out, it seems weird that a lot of the argument names are things like `arg0`; that makes me suspect you have some way of generating them, but it's not immediately clear to me what it is or how to redo it.

ctcisar:
Hi Adam,
Yes, I have the source file for the graphs (I made them in PowerPoint...). Should I add it to the repository or email it to you?
 
For the glossary, I sorted them manually and used sphinx to generate the text. I used the "function" specifier in autodoc, which imported from my local copy of odgi and automatically generated text.  My initial plan was to leave it like this, but because odgi isn't importable directly from the github repository it w doesn't generate the text on rtd. 

To get around this, I modified the python code in the autodoc module of sphinx to print the rst markdown to system output, and piped the output from `make html` to a text file. It's a bit hacky, but it was the method that I found to be the least manual work. [Method described here](https://stackoverflow.com/questions/2668187/make-sphinx-generate-rst-class-documentation-from-pydoc).

adamnovak:
I think if you could commit and PR the figure source files, that would be great.

For the glossary sphinx hack, you started with a file that asked for
the docs for each function in its appropriate place in the glossary,
which you then ran locally to generate the pre-baked RST with your
hacked Sphinx, which you then committed so that RtD could run on it.

Can you commit the original source file that you ran through your
hacked Sphinx? It looks like we could use
https://stackoverflow.com/a/31648880 and have a script of some sort to
redo what you did manually to generate the pre-baked RST file, but
only if we have the original source one that asks autodoc to run for
each method in its appropriate place in the glossary.

ekg:
This is great!

A technical note, for git and for our own future sanity. Please include GFA or text files rather than binary blobs in .odgi format. In the tutorial, have the user build the odgi graph from the GFA. If we change the format in the future, we'll need to make a ~200kb patch  to the repo. This can be awkward.

Sorry that I missed this until now.

ekg:
OK, maybe this is worse than I thought. I'm actually not able to load this file:

```
-> % odgi view -i docs/exdata/cactus-brca2.odgi
terminate called after throwing an instance of 'std::runtime_error'
  what():  error: Serialized handle graph does not match deserialzation type.
[1]    158975 abort (core dumped)  odgi view -i docs/exdata/cactus-brca2.odgi
```

So I can't render it into GFA. Would you please add another PR that removes this file and replaces it with the equivalent GFA?

ctcisar:
I submitted a PR with the original file that I started from-- it was a .vg file from [the Global Alliance Graph Bake-Off](http://public.gi.ucsc.edu/~anovak/graphs/).  

I included a small blurb on how to convert from another format, but for now I haven't found an "official" way to do it.  The way I made this file was by using VG to save it as a .json, and then manually read it into python and stored it as a odgi graph, which was saved with "serialize".  I am still able to load this into a python ODGI graph-- is there any reason why this wouldn't be compatible?
```
Python 3.6.9 (default, Nov  7 2019, 10:44:02)
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import odgi
>>> g = odgi.graph()
>>> g.load("../docs/exdata/cactus-brca2.odgi")
>>> g.get_node_count()
1134
```

ekg:
I see. The new magic number stuff isn't wrapped by the python bindings. So
the python serialized odgi files are incompatible.

I do think that having the GFA is better and more future proof.

adamnovak:
I don;t think the GFA can be loaded from Python using the odgi class,
can it? You would need the odgi command line tool to build the odgi
file from the GFA first.

I also don't understand why serialize/load as wrapped in the Python
bindings wouldn't include the magic number code; it looks to me like
they are provided by SerializableHandleGraph as they should be. Maybe
what happened was Cecilia made those files with the old version of
odgi that didn't have magic numbers yet, and then we merged in the PR,
and now the tutorial code won't work either because the serialize/load
behavior has changed under it. Cecilia, when you run the tutorial
code, are you running on your branch, or on master with your code
merged into it?

ekg:
Yes that seems right.

We can expose the load from GFA function in python. Right now the idea
would be to transform the graph with odgi from the command line and then
load it from python.

There is nothing wrong with building the graph bit by bit from a JSON
specification. That's kind of cool. I was just worried about holding onto
binary assets that might become hard to read later. Sorry if I didn't
explain myself.

Issue #78, opened 2020-02-06
Show nucleotide position in each bin for each path


josiahseaman:
In order to really make odgi and dependent visualizations be able to inter-operate with the rest of the Bioinformatics world, we need nucleotide positions for any feature, which will be different for each path.  Here's what I propose:

- [ ] Find code in odgi bin that is generating mean_pos
- [ ] Find the total length of each path
- [ ] Add center_nucleotide to Bin[3]
- [ ] Output center_nucleotide in bin output json
- [ ] Introduce JSON format version number so we can detect data incompatibility in the future

ekg:
Ah I see what could go wrong. We should add the center_position in parallel
so as to not break odgi bin visualization in R.

josiahseaman:
It shall be done as you say.

josiahseaman:
The latest PR is changed to include first_nucleotide and last_nucleotide.  No median.

subwaystation:
Is there still something open here @josiahseaman or can we close this?

PR #79, opened 2020-02-07
#78 add center_nucleotide output to bin json using the median of a list of positions

josiahseaman:
This gives the exact correct nucleotide position.  My only concern is that it might be slower than before.  However, if JSON output is not a significant portion of total runtime then that doesn't matter.

### DIFF
--- a/src/algorithms/bin_path_info.cpp
+++ b/src/algorithms/bin_path_info.cpp
@@ -3,13 +3,6 @@
 namespace odgi {
 namespace algorithms {
 
-int signum(double value){
-	/*Used for approximating medians without a list.  Caps values at -1,0, or 1 */
-	if(value < -1){ return -1;}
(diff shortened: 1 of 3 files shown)


josiahseaman:
It looks like the Travis build is failing because it can't get the code?
```
$ sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y
Error: retrieving gpg key timed out.
The command "sudo add-apt-repository ppa:ubuntu-toolchain-r/test -y" failed and exited with 1 during .
```

ekg:
Awesome!

I've tried to restart the build. Seems things have been flaky lately. This will get easier when Travis moves to Ubuntu 20.04.

josiahseaman:
Center_nucleotide and the surrounding median math was dropped in favor of the much simpler first_nucleotide and last_nucleotide output.

PR #80, opened 2020-02-18
Index bins

ekg:
going to open this to take a look

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,6 +267,7 @@ add_library(odgi_objs OBJECT
   ${CMAKE_SOURCE_DIR}/src/subcommand/layout_main.cpp
   ${CMAKE_SOURCE_DIR}/src/subcommand/flatten_main.cpp
   ${CMAKE_SOURCE_DIR}/src/subcommand/break_main.cpp
+  ${CMAKE_SOURCE_DIR}/src/subcommand/find_bin.main.cpp
   ${CMAKE_SOURCE_DIR}/src/algorithms/topological_sort.cpp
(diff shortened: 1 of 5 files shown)


subwaystation:
So far some preliminary results for t.gfa. Note that with increasing bin size, the size of the vector increases, too. Weird. Maybe the example is too small to actually show the compression effect.
[first_results.txt](https://github.com/vgteam/odgi/files/4225948/first_results.txt)

subwaystation:
Finally some results for our _A. thaliana_ data.
[sebastian.Athaliana.all.50000.gfa.og.w100000.eval.txt](https://github.com/vgteam/odgi/files/4236989/sebastian.Athaliana.all.50000.gfa.og.w100000.eval.txt)
`dac_default` would produce an index of ~40MB for Chr1 of one plant. This seems fine.
@ekg What is your impression?

Upscaling that to 3.000.000.000 bp for the human genome, we would have an index size of ~3.7GB.
So roughly the size of "phase" full human genome linear reference.

subwaystation:
I just took a look at the `vg` paper and as far as I got it, the index size for `xg` was not reported here.

ekg:
It isn't reported independently. I think I may have noted it in my thesis.
The HandleGraph paper in prep will provide results.

This seems rather large to me, at more than a byte per bp. It's simple
though.

subwaystation:
We can also try out the `xg` way. Should be adaptable pretty fast.

subwaystation:
@ekg I would close this request, as this was an experiment, only. But I would like to keep the branch just to show later, that this simple concept has certain limits. Thanks ;)

ekg:
That's fine, close if you'd like. I won't delete branch.

PR #82, opened 2020-02-27
Path index

subwaystation:
This is work in progress.


src/algorithms/xp.cpp
```diff
@@ -0,0 +1,724 @@
...
+        for_each_path_element([&](const std::string& path_name, const nid_t& node_id, const bool& is_rev, const std::string& cigar, const bool& is_empty, const bool& is_circular) {
+            if (path_name != curr_path_name && !curr_path_name.empty()) {
+                // build the last path we've accumulated
+                build_accumulated_path();
+                curr_path_steps.clear();
+                curr_is_circular = false;
+            }
+            curr_path_name = path_name;
+            if (!is_empty) {
+                handle_t visiting = get_handle(node_id, is_rev);
```
subwaystation: @ekg It seems that I need the `get_handle` function from https://github.com/vgteam/xg/blob/751f6c5552fc58055b0ca6295d132c116ed533c0/src/xg.cpp#L1629 in order to fill the path vectors. However, that implies, that I need to store `g`, `g_iv`, `g_bg_select` in XP. And these vectors must be filled and curated together with the edge vectors beforehand.
Which basically means, I have to create a full "XG" anyhow? Or do you see a way to circumvent that?

src/algorithms/xp.cpp
```diff
@@ -0,0 +1,166 @@
...
+    XP::~XP(void) {
+        // Clean up any created XPPaths
+        while (!paths.empty()) {
+            delete paths.back();
+            paths.pop_back();
+        }
+    }
+
+    /// build the graph from a graph handle
+    void XP::from_handle_graph(const PathHandleGraph& graph) {
```
subwaystation: @ekg This will be our conversation starter tomorrow ;)

src/algorithms/xp.cpp
```diff
@@ -0,0 +1,427 @@
...
+        sdsl::structure_tree::add_size(child, written);
+        return written;
+    }
+
+    void XP::deserialize_members(std::istream& in) {
+        // simple alias to match an external interface
+        load(in);
+    }
+
+    void XP::load(std::istream& in) {
```
subwaystation: @ekg when executing this method, I get
```
free(): double free detected in tcache 2
Aborted (core dumped)
```

src/algorithms/xp.cpp
```diff
@@ -0,0 +1,427 @@
...
+                in.unget();
+                in.unget();
+            }
+        } else {
+            // Put back the one character
+            in.unget();
+        }
+
+        try {
+            sdsl::read_member(pos_map_iv, in);
```
subwaystation: This and the following line are responsible for the problem. I just don't know why. @ekg

subwaystation: So only `sdsl::read_member(pos_map_iv, in);` is at fault....... hmmmm

subwaystation: So the problem was, when serializing the index to file, I had to serialize the pos_map_iv to disk instead of storing it as a member via sdsl.

src/unittest/pathindex.cpp
```diff
@@ -0,0 +1,76 @@
...
+                // TODO
+                /*
+                path_handle_t p_h_i = path_index.get_path_handle(f);
+                step_handle_t s_h_i = path_index.get_step_at_position(p_h_i, 12);
+                step_handle_t s_h_g = graph. // TODO @ekg I don't know how to do this.
+                REQUIRE(path_index.get_path_handle_of_step(s_h_i) == graph.get_path_handle_of_step(s_h_g));
+                */
+
+                // FIXME @ekg This does not run through.
+                // REQUIRE(as_integer(path_index.get_path_handle("5")) == as_integer(graph.get_path_handle("5")));
```
subwaystation:
```bash
/home/heumos/git/odgi/src/unittest/pathindex.cpp:49: FAILED:
REQUIRE( as_integer(path_index.get_path_handle("5")) == as_integer(graph.get_path_handle("5")) )
with expansion:
1 == 0
```

@ekg I am confused why this is the case. You have any idea?

src/algorithms/xp.cpp
```diff
@@ -0,0 +1,612 @@
...
+        if (occs.size() > 1) {
+            std::cerr << "error [xp]: multiple hits for " << query << std::endl;
+            exit(1);
+        }
+        if(occs.size() == 0) {
+            // This path does not exist. Give back 0, which can never be a real path
+            // rank.
+            return as_path_handle(0);
+        }
+        return as_path_handle(pn_bv_rank(occs[0])+1); // step past '#'
```
ekg: Like XG, you're adding 1 to the path handle here. This is why it's different than odgi. The difference is fine. We don't expect there to be stable handle definitions across different implementations.

subwaystation: I see :)
Thanks Erik!

subwaystation:
As @josiahseaman was happy with this, could you @ekg please merge it in?
I will continue the `node -> path_handles / step_ranks` on this branch, so please don't delete it. Thanks!

PR #85, opened 2020-03-04
Assembly tweaks

ekg:
Working on viral assembly problem, implementing graph cleaning and genome inference steps in odgi to work on variation graphs built directly from aligned reads.


ekg:
This isn't complete, but I'm merging it to document.

PR #89, opened 2020-03-20
CORS

subwaystation:
I need to set the correct headers in order to fix the CORS problem.


subwaystation:
So far it works, but let's not merge it in, yet.

PR #94, opened 2020-03-29
tsv based binning was broken beacuse the header wasn't updated

ekg:


subwaystation:
Shall we also add some metadata header lines?

We have that for JSON https://github.com/vgteam/odgi/blob/d31ed1baee0655f4def005f8e4127bde5c1fda5e/src/subcommand/bin_main.cpp#L89 but not for .tsv, yet. I did not fill it in, because I didn't know how you would process the .tsv downstream.
Proposition:
```
#odgi_version    number
#bin_width    number
#pangenome_length    number
```
@ekg

ekg:
In general tsv files just have one header line. Let's leave this for the
JSON format.

PR #96, opened 2020-03-30
Adding pure FASTA output for Pangenome #88

### LINKED ISSUE #88: Special case for bin_width=1 sequence output
At --bin_width=1 the json produced looks ridiculous.  For supporting Schematize, this is going to be a common use case.  Everytime we load in a graph with odgi we'll ask for bin width = 1, 10, 100, 1000, 10000, 100000 at minimum.  

Currently, the output would look like `{id=1, seq='C'}{id=2, seq=G}{id=3, seq='T'}`.  That's 43 characters for "CGT". In this special case, index in the string +1 is exactly the same as bin id, so they don't need to be listed.

New output should be:
`{pangenome_sequence="CGTACGTACGTACGTACTACTCAGCTAGCTAGCTACGTCGAGTCTTACTCTAGATC"}
{path_name="ATH17...`
No bin declarations should be included.  We'll also make a special case for bin_width=1 in schematize to look for the unique key "pangenome_sequence" which should only occur once in the file.  Besides this and the lack of bin declarations, the rest of the file can be the same including the bin_ids in the path traversals and links.

See also #49 #86

subwaystation:
@josiahseaman This brings an option `-f [FILE], --fasta [FILE]` to store the pangenome sequence in FASTA format in the specified FILE. Each sequence line will be at most 80 nucleotides long. The sequence header is the name of the FILE.

Please give me some feedback, if that suffices for what you had in mind.

### DIFF
--- a/src/algorithms/bin_path_info.cpp
+++ b/src/algorithms/bin_path_info.cpp
@@ -10,6 +10,7 @@ void bin_path_info(const PathHandleGraph& graph,
                                             const std::vector<std::pair<uint64_t, uint64_t>>&,
                                             const std::map<uint64_t, algorithms::path_info_t>&)>& handle_path,
                    const std::function<void(const uint64_t&, const std::string&)>& handle_sequence,
+                   const std::function<void(const string&)>& handle_fasta,
                    uint64_t num_bins,
                    uint64_t bin_width) {
     // the graph must be compacted for this to work
@@ -35,6 +36,8 @@ void bin_path_info(const PathHandleGraph& graph,
     for (uint64_t i = 0; i < num_bins; ++i) {
         handle_sequence(i+1, graph_seq.substr(i*bin_width, bin_width));
     }
+    // write out pangenome sequence if wished so
+    handle_fasta(graph_seq);
     graph_seq.clear(); // clean up
     std::unordered_map<path_handle_t, uint64_t> path_length;
     graph.for_each_path_handle([&](const path_handle_t& path) {
--- a/src/algorithms/bin_path_info.hpp
+++ b/src/algorithms/bin_path_info.hpp
@@ -34,6 +34,7 @@ namespace odgi {
                                                     const std::vector<std::pair<uint64_t, uint64_t>> &,
                                                     const std::map<uint64_t, algorithms::path_info_t> &)> &handle_path,
                            const std::function<void(const uint64_t &, const std::string &)> &handle_sequence,
+                           const std::function<void(const std::string&)> &handle_fasta,
                            uint64_t num_bins = 0,
                            uint64_t bin_width = 0);
     }
--- a/src/subcommand/bin_main.cpp
+++ b/src/subcommand/bin_main.cpp
@@ -100,7 +100,7 @@ int main_bin(int argc, char** argv) {
     std::function<void(const uint64_t&,
                        const std::string&)> write_seq_json
         = [&](const uint64_t& bin_id, const std::string& seq) {
-        if (args::get(write_seqs_not)) {
+        if (args::get(write_seqs_not) || fa_out_file) {
             std::cout << "{\"bin_id\":" << bin_id << "}" << std::endl;
         } else {
             std::cout << "{\"bin_id\":" << bin_id << ","
--- a/src/subcommand/bin_main.cpp
+++ b/src/subcommand/bin_main.cpp
@@ -3,6 +3,8 @@
 #include "args.hxx"
 #include "algorithms/bin_path_info.hpp"
 
+#include <regex>
+
 namespace odgi {
 
 using namespace odgi::subcommand;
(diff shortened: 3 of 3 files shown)


josiahseaman:
Great, Simon. I will take a look at this.  Is "sequence" still defined in every bin (in json) or does this replace that functionality?

josiahseaman:
Bin math: Bin 0 is a reserved meaning.   Biologists count nucleotide index starting at 1, not 0.  That means with bin width 1,000, position 1 -1,000 is in bin 1.  Target position X is in Bin label ceil(X /1,000).  Fasta position in file (0 indexed) for Bin Y = [(Y - 1) * 1,000 : Y * 1,000] non-inclusive of the last nucleotide (default behavior in Python and Javascript).   Nucleotide Index Z for Pangenome is Fasta[Z-1].

subwaystation:
If you don't set the `-s, --no-seqs` it would still output the sequences in the bins additionally. 
I can change it so that if you output to FASTA there won't be any sequences written in the bins.

subwaystation:
I can follow you until you start about FASTA position in file. Are you talking about a classical FASTA index like http://www.htslib.org/doc/faidx.html here?
Else, could you add a concrete example? Thanks!

josiahseaman:
@subwaystation Yes, I think that it should suppress "sequence" key in bins if you specify a FASTA output.  Personally, I'd make this the default behavior, but all we need is the option.  Please implement this check then we can merge.

ekg:
In odgi I'd prefer to use 0-based for everything. But did we settle on
bin=1-N?

BED files are 0-based. SAM files are 0-based. VCF files are 1-based.

It's all arbitrary, biologists use different schemes interchangeably.

subwaystation:
@ekg I somehow had the feeling you would step up in here :P

In odgi we still a have a 0-based system. I did not change that in bin. Still at 0 to N-1.

I understand Josiah's concerns about biologists being used to a 1-based system. Before I worked in odgi, I only saw 1-based files like VCF, GFF, etc.

So internally we can work 0-based. But user interaction can be 1-based. Does that make sense?

josiahseaman:
((Sorry for expanding scope on this issue. This PR can be merged.))
The only change I'd request is excluding the sequence from the Bin files #88.  I think this should automatically be excluded if you provide a FASTA path, otherwise there's redundancy.  I also think the fasta path should be included in the bin.json header along with our other things we've started putting in the header:
`{"odgi_version": 10,"bin_width": 100,"pangenome_length": 52441, "pangenome_file":"b1phi1-pangenome.fa"}`


Sorry to start a rabbit trail.  I don't think the indexing needs to be changed, I was just making notes for myself while making sure I understood the math for displaying the correct sequence in Schematize.

subwaystation:
@josiahseaman I added code regarding the automatic exclusion of the sequence in bin files.
I did not alter the header, because we only want to generate a FASTA once for each graph and not for every bin width. So sometimes the additional header would be there and sometimes not. Which seems like a nonstandardized way.

josiahseaman:
@subwaystation Would you like me to try and recruit another C++ developer to free up some of your time?  They could finish the auxiliary checks in your branch and then we could close out this PR.

josiahseaman:
Good call Simon!  You're correct.  I'm satisfied with that feature set.  I'll need to setup a test environment before I can follow proper procedure.

josiahseaman:
@subwaystation @ekg I'm testing and reviewing this change right now.  What is the test command and file you used?

josiahseaman:
Hey Simon, I'm really happy these changes have been made.  Thank you for all your hard work.  This change is going to mean we can show sequence in the Pantograph browser.

I've run tests in deNBI: /mnt/vol1/js/i88_B1phi1_review_fasta/ using b1phi1.gfa.  
Input was [run1.B1phi1.i1.seqwish.gfa.tar.gz](https://github.com/vgteam/odgi/files/4103860/run1.B1phi1.i1.seqwish.gfa.tar.gz).

Full script:
```bash
#!/bin/bash

ODGI=~/software/odgi/git/review/bin/odgi
GFA=run1.B1phi1.i1.seqwish.gfa
OG=${GFA%.gfa}.og
SOG=${GFA%.gfa}.sorted.og
FASTA=${GFA%.gfa}.fasta


## Build the sparse matrix form of the gfa graph
echo "### odgi build"
BLDPREF=${0%.sh}_01_build
/usr/bin/time -v -o ${BLDPREF}.time \
ionice -c2 -n7 \
$ODGI build \
--progress \
--gfa=$GFA \
--out=$OG \
> ${BLDPREF}.log 2>&1
#--sort \

## Sort paths by 1D sorting
echo "### odgi sort"
SRTPREF=${0%.sh}_02_sort
/usr/bin/time -v -o ${SRTPREF}.time \
ionice -c2 -n7 \
$ODGI sort \
--pipeline="bSnSnS" \
--sgd-use-paths \
--paths-max \
--progress \
--idx=$OG \
--out=$SOG \
> ${SRTPREF}.log 2>&1


echo "### odgi bin"
for w in 1 10; do
	BIN=${GFA%.gfa}.w${w}.json
	BINPREF=${0%.sh}_04_bin_w${w}
	/usr/bin/time -v -o ${BINPREF}.time \
	ionice -c2 -n7 \
	$ODGI bin \
	--json \
	--idx=$SOG \
	--bin-width=${w} \
	--fasta $FASTA \
	1> $BIN \
	2> ${BINPREF}.log &
done
echo "Done outputting bin and fasta."
## generate the index
#$ODGI pathindex \
#--idx=$SOG \
#--out=${SOG}.xp \
```
In particular, I removed the --no_seq argument and verified the .json no longer has sequence.  However, the bin declarations are still there, even though they have no information content.
```
{"bin_id":1}
{"bin_id":2}
{"bin_id":3}
{"bin_id":4}
{"bin_id":5}
{"bin_id":6}
{"bin_id":7}
```
@ekg would you mind these being omitted entirely? I believe they were created for the sole purpose of containing sequence.  A single header `{"odgi_version": 10,"bin_width": 1,"pangenome_length": 52441}` has all the information you need about the number of bins. For the bin1 output json I removed these and it reduced the output by 10x.

I'd also like the header to reference the "fasta_file" name.  Still, all the work Simon has done is a tangible improvement in the code and I believe it should be merged immediately.

subwaystation:
I missed the new related issue. Opening again.

PR #101, opened 2020-04-08
Command-line option for removing gap links from the output

lomereiter:
See https://github.com/vgteam/odgi/issues/48 for the background.

### DIFF
--- a/src/algorithms/bin_path_info.cpp
+++ b/src/algorithms/bin_path_info.cpp
@@ -12,7 +12,8 @@ void bin_path_info(const PathHandleGraph& graph,
                    const std::function<void(const uint64_t&, const std::string&)>& handle_sequence,
                    const std::function<void(const string&)>& handle_fasta,
                    uint64_t num_bins,
-                   uint64_t bin_width) {
+                   uint64_t bin_width,
+                   bool drop_gap_links) {
     // the graph must be compacted for this to work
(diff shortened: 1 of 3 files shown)


ekg:
This looks sharp. The debugging information will be printed for every path, so it might blow up for some graphs. But otherwise good to merge in.

lomereiter:
It's the sum for all paths that's printed. Indentation is confusing in some places, this codebase would benefit from a clang-format run.

ekg:
I wasn't familiar with clang-format. I'll take a look.

josiahseaman:
@6br Do you think we should switch to using this option flag in our pipeline?

6br:
In general, because we can see the gap as just an absence of paths on the component, it would be better for our pipeline to turn on the `drop_gap_link` option.

Still, I am not quite confident in the following case if we remove all gap links.
Imagine the situation that the left of the gap is the end of the path, and the right of the gap is the start of the path, due to rearrangement. There is no actual link across the gap. We might confuse in such a case whether there is a hidden gap _link_ or not. We need to know where the path starts or ends. I recommend waiting for https://github.com/graph-genome/Schematize/issues/2.

lomereiter:
@6br is your concern specifically about the [0, start] and [end, 0] links? This PR is based on the Python reference implementation in component_segmentation, so your concern also applies there.

6br:
Yes this is one of my concern.

PR #102, opened 2020-04-08
I99 odgi bin nuc pos

subwaystation:
This PR adds the detailed nucleotide position for each bin.

### DIFF
--- a/src/algorithms/bin_path_info.cpp
+++ b/src/algorithms/bin_path_info.cpp
@@ -72,11 +72,11 @@ void bin_path_info(const PathHandleGraph& graph,
                         bins[curr_bin].mean_pos += path_pos++;
                         nucleotide_count += 1;
 						if((bins[curr_bin].ranges.size() == 0) ||
-						    ((nucleotide_count -  bins[curr_bin].ranges[bins[curr_bin].ranges.size() - 1].second) > 1)){
+						    ((nucleotide_count -  bins[curr_bin].ranges.back().second) > 1)){
 						    bins[curr_bin].ranges.push_back(std::make_pair(nucleotide_count, nucleotide_count));
 							// bins[curr_bin].first_nucleotide = nucleotide_count;
 						} else {
-                            bins[curr_bin].ranges[bins[curr_bin].ranges.size() - 1].second = nucleotide_count;
+                            bins[curr_bin].ranges.back().second = nucleotide_count;
 						}
                         last_bin = curr_bin;
                         last_is_rev = is_rev;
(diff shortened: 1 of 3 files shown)


subwaystation:
I only modified the JSON output. 

The .tsv output will still be the same, so it does not break scripts in R, etc.

@ekg Please merge, if you think it's fine. Thanks!

josiahseaman:
Have you profiled the change in run speed and output size?  I had almost identical code in odgi earlier and I stripped it out because it seemed we only needed sign post nucleotides (first, last, median), not every single one. I probably missed a conversation. Which feature are you trying to support on the front end?

subwaystation:
Sorry, I should have specified that. So in order to bring in annotation in our view, we either need a `bin -> node_id` mapping for each path or a `bin -> nuc_pos` mapping. I discussed this with @JervenBolleman . Else we would not know, which feature corresponds to which position. 
If someone is aware of a better solution, I am happy to go down another road.

src/algorithms/bin_path_info.cpp
```diff
@@ -71,10 +71,13 @@ void bin_path_info(const PathHandleGraph& graph,
                         }
                         bins[curr_bin].mean_pos += path_pos++;
                         nucleotide_count += 1;
-                        if(bins[curr_bin].first_nucleotide == 0){
-                            bins[curr_bin].first_nucleotide = nucleotide_count;
-                        }
-                        bins[curr_bin].last_nucleotide = nucleotide_count;
+						if((bins[curr_bin].ranges.size() == 0) ||
+						    ((nucleotide_count -  bins[curr_bin].ranges[bins[curr_bin].ranges.size() - 1].second) > 1)){
```
lomereiter, line 75: https://en.cppreference.com/w/cpp/container/vector/back

lomereiter:
> Have you profiled the change in run speed and output size? I had almost identical code in odgi earlier and I stripped it out because it seemed we only needed sign post nucleotides (first, last, median), not every single one.

AFAIU this PR only changes a single range [first, last] to multiple, to account for loops which would have otherwise be lumped into [ranges[0].first, ranges[N].last]. It's not dumping "every single one".

src/algorithms/bin_path_info.cpp
subwaystation: I was not aware of this, will change accordingly.

subwaystation:
@dimatr If I merge this in, we would have to update the JSON parsing in CS.
So I am going to wait until we have a similar PR in CS to update synchronously.

subwaystation:
We will need odgi_version = 11

subwaystation:
Following @josiahseaman Version 12 bumps merging this in, too.

PR #104, opened 2020-04-25
trade space for time by storing edges in their own hacked vector

ekg:
This adjust the node data model to have a string for the node sequence, a hacked vector for the edges, and another for the path steps.

It will take more memory, but it also removes a few bytes used for recording the length of the various parts of each node.

### DIFF
--- a/src/node.cpp
+++ b/src/node.cpp
@@ -5,55 +5,33 @@
 namespace odgi {
 
 uint64_t node_t::sequence_size(void) const {
-    return seq_bytes();
+    return sequence.size();
 }
 
-const std::string node_t::sequence(void) const {
-    const std::string res((char*)bytes.data()+seq_start(), seq_bytes());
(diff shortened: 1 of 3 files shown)


jeizenga:
Hey, if you're doing engineering on the odgi graph, I'll once again suggest that we make libbdsg a dependency of odgi so that we can keep the implementations from going out of sync. Otherwise, the odgi graph implementation in libbdsg is going to go stale, which is especially not great since we just submitted a preprint, and people are going to start paying attention to it soon.

ekg:
I agree. I was going to do this and then start flipping out different
libbdsg implementations.

ekg:
I'd like to use libbdsg, but the repo has become very large and I'm unable to build it due to pybind11 configuration problems. I see that I have to run a complex build process to update the python bindings.

What's going on there?

PR #106, opened 2020-04-26
Python objects should implement  hash and equals methods.

JervenBolleman:
Also make step and edge information easier to access.

### DIFF
--- a/src/pythonmodule.cpp
+++ b/src/pythonmodule.cpp
@@ -14,14 +14,40 @@ namespace py = pybind11;
 PYBIND11_MODULE(odgi, m)
 {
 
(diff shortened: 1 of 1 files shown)


ekg:
Should this be merged?

JervenBolleman:
The hash and equals code part yes. The other. Part is nice to have. Do double check that it still applies, it's been a while since I opened the pull request.

JervenBolleman:
Yeah this is actually important and breaks lot's of usecases and simple things like sets in python

ekg:
I'll close this as it's gone stale. Reopen if we should pick this back up.

PR #107, opened 2020-04-27
Docker file with a static odgi build in

JervenBolleman:
Add a docker build. It's a huge image but we can work on that in tree.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,0 +1,31 @@
+FROM debian:buster
+ENV DEBIAN_VERSION=buster-20200327
+ENV ODGI_COMMIT=dc76da391943a6bd6a7784e42f946dc478b865a5
+
+WORKDIR "/var/lib/"
+RUN DEBIAN_FRONTEND=noninteractive \
+    && apt-get update -y \
+    && apt-get install -y --no-install-recommends \
+      build-essential \
+      ca-certificates \
+      cmake \
+      git \
+      python3-dev \
+      python3-setuptools \
+      zlib1g-dev \
+    && rm -fr /var/lib/apt/lists/* \
+    && update-ca-certificates \
+    && git clone --shallow-submodules --recursive https://github.com/vgteam/odgi.git \
+    && cd odgi \
+    && git checkout "${ODGI_COMMIT}" \
+    && cmake -DBUILD_STATIC=1 -H. -Bbuild \
+    && cmake --build build -- -j $(nproc) \
+    && rm -rf deps \
+    && rm -rf build \
+    #&& rm -rf lib/*.a \
+    && ls -s bin/odgi /usr/bin \
+    && apt-get remove -y --autoremove  build-essential cmake \
+    && apt-get clean
+ENV PATH="/var/lib/odgi/bin/:${PATH}"
+ENTRYPOINT ["odgi"]
+#CMD ["bash"]


subwaystation:
Maybe @6br 's https://github.com/graph-genome/pipeline/blob/master/Dockerfile would present a slimmer alternative?

JervenBolleman:
@subwaystation that is at least a GB larger

my podman images gives during the build already :(
```
localhost/jerve/pipeline         latest    2487a5ae16a5   34 seconds ago      3.61 GB
localhost/jerven/odgi             latest    a8664318a7a2   39 minutes ago      1.05 GB
```

subwaystation:
@JervenBolleman Damn! Worth a try I thought.

JervenBolleman:
To note even just the odgi parts are already larger in @6br example. Just from leaving in all the build tools. 

Removing odgi/lib/libodgi.a saves 334mb and odgi test passes without that.

6br:
FYI: https://quay.io/repository/biocontainers/odgi?tab=tags is a docker image with odgi, just 163.7 MB!

subwaystation:
Right, I forgot about the Bioconda release docker. But we don't have access to the `Dockerfile` itself?

6br:
Unfortunately, I have not found Dockerfile itself. At least we can glance at the Docker container's layers using this tool https://github.com/wagoodman/dive.

JervenBolleman:
Can't make it that small. As we will need the python library part which is 134mb. But taking some other steps get's it down to 704mb. Of which 321 mb is odgi.

JervenBolleman:
It's now 550mb. 230mb is not odgi. of which 110mb is python3.7 leaving 220mb mostly unaccounted for.

Having the python libs there is important so that spodgi and graph-genomecode can use this as a layer.

docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    && cmake --build build -- -j $(nproc) \
+    && rm -rf deps \
+    && rm -rf .git \
+    && rm -rf build \
+    && rm -rf lib/*.a \
+    && ls -s bin/odgi /usr/bin \
+    && apt-get remove -y --autoremove  build-essential cmake git \
+    && apt-get clean \
+    && apt-get purge \
+    && rm -rf /var/lib/apt/lists/* /tmp/*
```
mr-c, line 32: If you use build stages, then all this deleting isn't necessary

Here's my minimal image for odgi that demonstrates this technique in 201 MiB
```
FROM debian:stable-slim
WORKDIR /usr/src/app
RUN apt-get update && apt-get install -y git bash cmake make g++ python-dev && rm -rf /var/lib/apt/lists/*
RUN git clone --recursive https://github.com/vgteam/odgi.git
RUN cd odgi && cmake -H. -Bbuild && cmake --build build -- -j $(nproc) && cd build && make install
FROM debian:stable-slim
RUN apt-get update && apt-get install -y libgomp1 && rm -rf /var/lib/apt/lists/*
COPY --from=0 /usr/local/bin/odgi /usr/local/bin/
```

JervenBolleman: @mr-c thanks, that was a very helpful hint about build stages.

I made two post build stages, one as above.
The second with the minimal python3 install for downstream users using python3.

@subwaystation do you need both the binary and the python3 libs?

subwaystation: Until now, we would only need the binary for our workflow. I don't expect we will need the bindings in the future. Fingers crossed.

But, maybe @Flavia95 would have interest in a docker including Python3 libs?

JervenBolleman:
Test builds available at
https://hub.docker.com/repository/docker/jerven/odgi
and
https://hub.docker.com/repository/docker/jerven/odgi-python3

docker/Dockerfile
Flavia95: Wow, thanks @subwaystation.
That would be great for me. It would make my job much easier.

subwaystation:
The odgi docker works for me, all tests ran through. Thanks @JervenBolleman!
I also pulled and ran odgi-python3 and there did not seem any flaws. Could not really test it, though.

@Flavia95 Could you please give some feedback, if the odgi-python3 docker works for you? Thanks.

@JervenBolleman Do you want to use the Python3 docker in Spodgi, too? As a base image for it? Or how do you connect there?

JervenBolleman:
@Flavia95  can you try the [odgi-python3](https://hub.docker.com/repository/docker/jerven/odgi-python3) build here? and tell me if it includes enough? e.g. [spodgi dockerfile](https://github.com/pangenome/spodgi/blob/master/docker/Dockerfile)

JervenBolleman:
@ekg I have been using podman to build this as that doesn't need root. But other image builders should work as well.

Flavia95:
For now both work well.

For https://hub.docker.com/repository/docker/jerven/odgi-python3 I did: 
docker pull jerven/odgi-python
docker run -it -v /:/directoryofscript/ jerven/odgi-python3  and when import odgi works. 

The same if I run script.
Thank you.

subwaystation:
@JervenBolleman How about an automatic building process as described in https://docs.docker.com/docker-hub/builds/?
The process would be bound to one account, though.

subwaystation:
I can take care of it, I just want to know, if we should do this via e.g. my account or yours or @ekg 's or we create a shared one?

JervenBolleman:
> I can take care of it, I just want to know, if we should do this via e.g. my account or yours or @ekg 's or we create a shared one?

I would prefer an organization. odgiteam should be open at dockerhub. pangenome would be ok as well. For releases I would like a little bit more managed than only latest. Also for the ease of mind of @ekg. I think he fearlessly wants to  be able to hack on master and know that end users have good images to work with.

mr-c:
By either upgrading to a newer cmake and adding `--strip` to a `cmake --install` invocation, or manually calling `strip /usr/local/bin/odgi` I got my container down to 80.7MiB.

Here's the complete `Dockerfile` I'm using now:

```
FROM debian:bullseye-slim
WORKDIR /usr/src/app
RUN apt-get update && apt-get install --no-install-recommends -y \
  ca-certificates \
  bash \
  cmake \
  git \
  g++ \
  make \
  python-dev \
  && rm -rf /var/lib/apt/lists/*
RUN git clone --recursive https://github.com/vgteam/odgi.git
RUN cd odgi && cmake -DCMAKE_BUILD_TYPE=Release -H. -Bbuild && \
  cmake --build build --config Release -- -j $(nproc) && \
  cmake --install build/ --config Release -v --strip
FROM debian:bullseye-slim
RUN apt-get update && apt-get install -y libgomp1 && rm -rf /var/lib/apt/lists/*
COPY --from=0 /usr/local/bin/odgi /usr/local/bin/
```

JervenBolleman:
I am not sure that stripping the binary is ideal. Regarding bug reports.

mr-c:
@JervenBolleman I have no opinion, one way or the other; just wanted to make sure that those interested in a smaller container were aware

JervenBolleman:
@mr-c thanks for all the ideas. Best way to learn docker for me ;) 
@ekg we now have 3 docker images:

1. odgi (76mb)
2. odgi-debug (203mb)
3. odgi-python3 (251mb)

and one script to help build all 3 at once.

subwaystation:
I would vote for `pangenome`. We could use that account also for related projects like SpOdgi or the Pantograph pipeline. `Odgiteam` would be perfectly fine, too. Please give some feedback, which would fit you best @ekg @6br @mr-c 
@JervenBolleman We can have an automated build triggered by each push to master. Additionally, for each release, we should be able to override the default docker build process to build images for each release and tag. See https://docs.docker.com/docker-hub/builds/advanced/ and https://dev.to/samuelea/automate-your-builds-on-docker-hub-by-writing-a-build-hook-script-13fp. So Erik can hack on master without any drawbacks. As I am not experienced in such processes, it might take some time to set this up. Or would someone else volunteer?

6br:
I also agree to create an organization on dockerhub.

subwaystation:
We publish a docker image that actually contains odgi. Latest builds can be found at https://github.com/orgs/pangenome/packages.

PR #125, opened 2020-06-05
Sync up with libbdsg and backport improvements

adamnovak:
This backports a bunch of libbdsg ODGI improvements (like the hidden handle bugfixes, and the end-exclusive rewrite semantics) to odgi. It makes odgi file-compatible with libbdsg as of libbdsg https://github.com/vgteam/libbdsg/pull/70.

I need this so that I can update the tutorial at pangenome.github.io to work with the libbdsg Python bindings. Making the Python bindings share method names is easy, but getting the odgi tool's generated graphs to work with the library ODGI is kinda hard, so I have two big PRs that synchronize things.

I think we need this (or something that does the same thing) before we can resubmit the handle graph paper. Otherwise we'll be telling people we have ODGI bindings, but they won't work with odgi.

Unfortunately, this ends up undoing #122. I don't think #122 actually works as merged; it zeros out steps, but then the zeroed-out steps still count towards `get_step_count(handle_t)` on the handle they're on, which isn't right. If I change that, then the code to add new steps (which just uses `get_step_count()` to try and find the next step index to fill) will be wrong. I've added somne machinery to try and prepare for the path end markers changing, and path handle 0 being reserved as nonexistent, but the actual lazy delete is gone.

@ekg Can you take a look at this and see if it breaks anything I haven't noticed? It appears to pass all the unit tests, but I'm not really an odgi code expert.


ekg:
The lazy delete of steps is unfortunately essential to ongoing development
here. Can we correct the problem you found rather than undo #122? This was
doing what I expected in my tests, but you seem to have found a bug with
it. The step count call was incorrect, but iteration over steps and path
steps should not have been affected. I can add further automated tests to
verify that things are correct.

Does this synchronize the semantics for rewrite segment? If not, I can
address that too.

Thanks for this. It's very helpful. The path rewriting issue has been a
consistent problem for algorithms I've been writing. The smart thing is to
use the optimize call as a garbage collection cycle for deleted path steps
as is done for deleted handles.

ekg:
The changes here are kind of enormous, and it will take some time for me to review them. Would you please describe what's you mean to have altered?

Would you please update the PR to undo the change of the class name?

The object name has been graph_t. It's in the odgi namespace, so externally it's odgi::graph_t, which should reduce confusion enough in libbdsg. You can always do a typedef there to use ODGI.

ekg:
I think more thought is needed here. It would be nice to talk with you about it when we can, perhaps on the call tomorrow. 

The hidden node concept hasn't proven very useful, although in principle it could help when doing various cleaning operations on the graph. These operations would have to be supported in libhandlegraph to make sense here. For instance, we would need to know when a path step was on a deleted node. It's not clear that these semantics are helpful. Perhaps would be preferable to rewrite paths into pieces.

I'm also concerned that the overhead of checking the hidden node set for every interaction with a handle will be costly, but maybe this is efficient when the set is empty?

I need to review all the other function-level changes.

Otherwise, it's clear to me that integration tests are needed to confirm that there isn't a problem introduced by these changes (or what existing ones they resolve). I've been running odgi like a single person project, which has been OK, but isn't sustainable.

ekg:
We should rebase this on master, incorporate the bugfixes here, and if desired merge downstream into libbdsg.

As noted, a few things need to be changed:

- The object name shouldn't change. Libbdsg should match what's used here. 
- The deleted handle concept needs to be worked out completely in libhandlegraph, so that we can understand which path steps are on deleted nodes. The semantics aren't clear to me now. This is something I started but haven't replied on completely outside of lazy deletion of nodes.
- Path steps should be deleted lazily and reclaimed using optimize. 
- We need integration tests to confirm the behavior of command line tools. 

I may reopen here to implement some of these changes.

adamnovak:
We talked about this this morning with @jeizenga and decided we're going to try and backport fixes from libbdsg with single-issue PRs until all the features/fixes are in, and then snapshot stable odgi implementations over from odgi to libbdsg, possibly with some git subtree magic. This is all going to happen after the handlegraph paper resubmission, and we just won't use the pangenome.github.io docs in the paper.

* We can keep the `odgi::graph_t` name for the class in both places, and typedef as `bdsg::ODGI`, provided nobody tries to include libbdsg and pieces of odgi in the same downstream project, which probably will never happen.

* We can remove the deleted/hidden handle feature, as the odgi tool doesn't really use it and the handle graph API doesn't officially support it.

* We'll test/fix up the path step reclaiming semantics in this repo until we're happy with them.

* I'm not sure what to do about the integration tests for the odgi binary. We should have them if we're going to be making a bunch of revisions to the core data structure, but I don't like the idea that we can't change anything until we have some unspecified sufficiency of them because that seems like an easy way to stall ourselves. Plus, good unit tests for the graph should protect us from the same sorts of problems in this case. Maybe the right answer is to grab the Bash-TAP system from vg and start by testing some basic set of example commands?

ekg:
We can still reference the pangenome.github.io docs in the paper, in addition to the libbdsg ones. We might need a better explanation of the way they all relate, but it is going to be OK.

Keep in mind that @flavia95 and @subwaystation worked on these. They cover material that's relevant to the users of odgi, which is our only full interface to a pure dynamic/compact handle graph.

PR #127, opened 2020-06-12
resolve bug with multithreaded path-SGD, improve documentation, change seed

ekg:


ekg:
@subwaystation I've found a problem. Sorting certain kinds of graphs without edges causes an infinite loop. See these.
[V-352962.gfa.txt](https://github.com/vgteam/odgi/files/4770929/V-352962.gfa.txt)
[DRB5-3127.gfa.txt](https://github.com/vgteam/odgi/files/4770930/DRB5-3127.gfa.txt)

subwaystation:
@ekg Thanks fixing the bug so multithreaded path-SGD works again!

I took a look at the graphs you mentioned. I guess they either don't have enough nodes for path-SGD to work or they already present a perfectly sorted graph. Therefore, `delta_max` becomes 0 immediately, which was not handled by the current implementation.
This is fixed now.

Would be cool, if you could check again. For me it worked. 

PS: We really need some tests for the binary so we won't need the double checking.

subwaystation:
The default SGD also doesn't like your edge cases:

```
(gdb) run sort -i DRB5-3127.gfa.og -o DRB5-3127.gfa.sorted.og -P -p S -t 2
Starting program: /usr/bin/odgi sort -i DRB5-3127.gfa.og -o DRB5-3127.gfa.sorted.og -P -p S -t 2
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".
[New Thread 0x7ffff7a15700 (LWP 56356)]
[New Thread 0x7ffff7214700 (LWP 56357)]
[New Thread 0x7ffff6a13700 (LWP 56358)]

Thread 3 "odgi" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff7214700 (LWP 56357)]
0x00007ffff7beefad in pthread_mutex_trylock () from /usr/lib/libpthread.so.0
```
Will take a look.

subwaystation:
@ekg This is ready to merge.

PR #128, opened 2020-06-12
odgi viz can work in binned visualization mode (new options -w/--bin-width and -g/--no-gap-links)

AndreaGuarracino:
A more rigorous / theoretical way to test the functioning would be useful. If confirmed, I will update the documentation.

>Bob
ACGTAGCTGATCGCTGTCAGCGATCGAGCTGACTGCATCTAGCTTTTTTTTTTTTTTTTTTTGACGTCAGTAC
>Jorg
ACGTAGCTGATCGCTGTCAGCGATCGAGCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCGACTGCATCTAGCTTTTTTTTTTTTTTTTTTTGACGTCAGTAC
>Simon
ACGTAGCTGATCGCTGTCAGCGATCGAGCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCGACTGCATCTAGCCGACGTCAGTAC

Little example 1:
- visualized with Schematize (Simon, Jorg, Bob)
image

- odgi viz output with/without gap links
- image

Little example 2:
- vsualized with Schematize (Bob, Simon, Jorg)
image

- odgi viz output without/with gap links
image


AndreaGuarracino:
More complex example using an old pangenome produced by Public Sequence Resource with 1190 SARS-CoV-2 genomes.

The problem is the computational time, which passes from about 1,65 seconds to about 3,78 second.

with gap links / without gap links

image

AndreaGuarracino:
Mock examples. All the images are with/without gap links.

The gap link is not displayed.
> H	VN:Z:1.0
> S	1	AAAAAAA
> S	2	AAAA
> S	3	AAAAAAAAAAAAAA
> P	Path	1+,3+	*
> P	1	1+	*
> P	2	2+	*
> P	3	3+	*
image (1)

Links connecting nodes in different orientation are always displayed right now.
> S	1	AAAAAAA
> S	2	AAAA
> S	3	AAAAAAAAAAAAAA
> P	Path	1+,3-	*
> P	1	1+	*
> P	2	2+	*
> P	3	3+	*
image (2)

> H	VN:Z:1.0
> S	1	AAAAAAA
> S	2	AAAA
> S	3	AAAAAAAAAAAAAA
> P	Path	1-,3+	*
> P	1	1+	*
> P	2	2+	*
> P	3	3+	*
image (3)

> H	VN:Z:1.0
> S	1	AAAAAAA
> S	2	AAAA
> S	3	AAAAAAAAAAAAAA
> P	Path	1-,3-	*
> P	1	1+	*
> P	2	2+	*
> P	3	3+	*
image (4)

Here there is the node 2 between node 1 and 3, so the link connecting them is displayed.
> H	VN:Z:1.0
> S	1	AAAAAAA
> S	2	AAAA
> S	3	AAAAAAAAAAAAAA
> P	Path	1+,3+,2+	*
> P	1	1+	*
> P	2	2+	*
> P	3	3+	*
image (5)

> H	VN:Z:1.0
> S	1	AAAAAAA
> S	2	AAAA
> S	3	AAAAAAAAAAAAAA
> P	Path	1+,3+,1-	*
> P	1	1+	*
> P	2	2+	*
> P	3	3+	*
image (6)

josiahseaman:
This is really great work @AndreaGuarracino ! Well done.
image
In this example, I'm wondering how the lower paths ever get to the rearrangement.  I think this is probably from Schematize before "color by copy number" and this CCCCCCCCCC region is copy number 2?

AndreaGuarracino:
Thank you @josiahseaman, but it is still a work in progress. About your question, may the corrisonding GFA help you to have a clearer vision?

```
H	VN:Z:1.0
S	1	CCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCCC
L	1	-	2	-	0M
L	1	+	4	+	0M
S	2	ACGTAGCTGATCGCTGTCAGCGATCGAGC
L	2	+	3	+	0M
S	3	T
L	3	+	4	+	0M
S	4	GACTGCATCTAGC
L	4	+	6	+	0M
L	4	+	5	+	0M
S	5	TTTTTTTTTTTTTTTTTTTTT
L	5	+	7	+	0M
S	6	C
L	6	+	7	+	0M
S	7	GACGTCAGTAC
P	Bob	2+,3+,4+,5+,7+	*
P	Simon	2+,1+,4+,6+,7+	*
P	Jorg	2+,1+,4+,5+,7+	*
```

ekg:
This is looking good!

ekg:
What's the binned mode doing? Can you set the number of pixels per bin?

Right now, it's like we're binning at 1px, based on the width that's given with -x.

AndreaGuarracino:
It is a display mode very similar to that of Schematize. If at least one of these two new arguments is specified (`--bin_width` or `--drop-gap-links`), the image is created after a binning phase like in odgi bin. In this mode, the displayed links connect the bins, therefore they changes according to the bin_width. If only `--drop-gap-links` is specified, bin_width is calculated from the x-width. In binning mode, the images have always 1 bin per pixel (dangerous for large pangenomes and lower bin_width values).

I was discussing with @subwaystation on how to proceed and whether to confirm this behavior, maybe I was still not aware of something else.

ekg:
This seems doable. We can work with it to see if the API makes sense. Through documentation we can make clear how to use it, but we'll need some experience with it to feel out if it makes sense.

AndreaGuarracino:
A proposal to simplify the utilization for the users could be to introduce a third argument `-b/--binning-mode`:

`odgi viz -i graph.og -o image.png --binning-mode`
The bin_width will be calculated from the default x-width.

`odgi viz -i graph.og -o image.png --binning-mode -x 10000`
The bin_width will be calculate from the specified x-width.

`odgi viz -i graph.og -o image.png --binning-mode -x 10000 -w 1`
The x-width will be calculated from the specified bin_width (it will be set as the number of bins), warning the user if the value is bigger than X (something like 50k/100k pixels).

The `-g/--no-gap-links` works only in binning mode for hiding the gap links, returning an error if it is used without specifying the `--binning-mode` argument.

subwaystation:
@AndreaGuarracino This looks really good!

Played around with it a little bit, there is a major difference between the default mode and the binning mode regarding the drawing of links.
DOB-3112 fa gz gfa clean gfa og R x20
DOB-3112 fa gz gfa clean gfa og R b w200 g
Please draw all links out so we can see them clearly.

AndreaGuarracino:
Here you are

- on the left: `odgi viz ... -x 30`
- on the right: `odgi viz ... -x 30 -b`

image

src/subcommand/viz_main.cpp
```diff
@@ -534,10 +542,7 @@ namespace odgi {
                 uint64_t min_x = std::numeric_limits<uint64_t>::max();
                 uint64_t max_x = std::numeric_limits<uint64_t>::min(); // 0
 
-                // In binned mode, the min/max_x values changes based on the bin width
-                if (!args::get(binned_mode)) {
-                    _bin_width = 1;
-                }
+                // In binned mode, the min/max_x values changes based on the bin width; in standard mode, _bin_width is 1, so nothing change here
```
subwaystation: I think this is clear from your documentation above. We can delete this.

AndreaGuarracino:
Hi @ekg, I post some examples of the new binned mode so that you can judge if it could be a useful feature to include in odgi viz. Obviously, the advantage of such a mode strongly depends on the specific pangenome and sorting. Each image shows the bottom part of each odgi viz output, zooming in the links.

For the first example I used the `run1.B1phi1.i1.seqwish.gfa` file which @subwaystation passed me.

```
time odgi viz -i run1.B1phi1.i1.seqwish.sorted.og -o run1.B1phi1.i1.seqwish.sorted.binned.x1000.png -x 1000

real	0m0,940s
user	0m0,868s
sys	0m0,072s
```

In binned mode, if the bin width is not specified, it is calculated from the image width.
```
time odgi viz -i run1.B1phi1.i1.seqwish.sorted.og -o run1.B1phi1.i1.seqwish.sorted.binned.x1000.binned.png -x 1000 -b
Binned mode
bin width: 52.441
image width: 1000

real	0m0,945s
user	0m0,869s
sys	0m0,076s
```

In binned mode, when the `-g` is specified, it is printed the number of gap links removed and the total number of links.
```
time odgi viz -i run1.B1phi1.i1.seqwish.sorted.og -o run1.B1phi1.i1.seqwish.sorted.binned.x1000.binned.nogap.png -x 1000 -b -g
Binned mode
bin width: 52.441
image width: 1000
Gap links removed: 4583 of 6198 total links

real	0m0,947s
user	0m0,871s
sys	0m0,076s
```

run1 B1phi1 i1 seqwish sorted binned x1000 links_comparisonpng

In the second example I used again the pangenome produced by Public Sequence Resource with 1190 SARS-CoV-2 genomes.

```
time odgi viz -i relabeledSeqs_dedup_relabeledSeqs_dedup.sorted.og -o relabeledSeqs_dedup_relabeledSeqs_dedup.sorted.x5000.png -x 5000

real	0m3,868s
user	0m3,656s
sys	0m0,212s
```

```
time odgi viz -i relabeledSeqs_dedup_relabeledSeqs_dedup.sorted.og -o relabeledSeqs_dedup_relabeledSeqs_dedup.sorted.x5000.binned.png -x 5000 -b
Binned mode
bin width: 25.9358
image width: 5000

real	0m3,989s
user	0m3,657s
sys	0m0,332s
```

```
time odgi viz -i relabeledSeqs_dedup_relabeledSeqs_dedup.sorted.og -o relabeledSeqs_dedup_relabeledSeqs_dedup.sorted.x5000.binned.nogap.png -x 5000 -b -g
Binned mode
bin width: 25.9358
image width: 5000
Gap links removed: 401809 of 406254 total links

real	0m4,388s
user	0m4,072s
sys	0m0,316s
```

relabeledSeqs_dedup_relabeledSeqs_dedup sorted x5000 links_comparison

src/subcommand/viz_main.cpp
AndreaGuarracino: The comment is related to the `_bin_width` variable that I added to the denominators of the following lines of code, but if you click on the last commit on git, those lines are not visible and the comment seems unnecessary.

subwaystation: I somehow missed that, sorry! So all good.

subwaystation:
Hi @AndreaGuarracino I think these are great results, especially looking at the PubSeq data! There the gap link removal really shines, because the pangenome is mostly linear! Thanks for your hard work.

ekg:
Nice work @AndreaGuarracino!

The last examples you gave really drive home why this makes sense. In the future, we can think about making it the default behavior.

ekg:
Another viz trick would be to remove singleton fragments of the graph, where only one (or < N) paths are crossing the nodes. That would further compress things and make it even more MSA-like in these linear cases.

AndreaGuarracino:
Happy to know that. I am glad to give a little hand.

@ekg, regarding your suggestion on singleton nodes, given the following scenario where the third node is only in one path
example_singleton

do you suggest hiding the singleton node?
example_singleton hide

or do you suggest removing it as if it were not present in the graph at all?
example_singleton remove

The second case would lead to a compression in the visualization, but perhaps it could be misleading. To avoid confusion, we could think of a vertical line, or something else, to indicate where 'a cut' has occurred in the visualization..

ekg:
The idea was to hide it as if it isn't in the graph at all. But, this introduces the problems you're describing.

The general way to do this would be to make odgi prune be capable of retaining path information for cases when we remove low-depth nodes. There was an idea to do this with "hidden" nodes that retained information about paths, but I removed that due to performance issues. In #125, there is a partial completion of this idea, but I'm not sure it makes sense in the current form.

The problem I ran into was confusion about the semantics of the paths that were divided by removal of low-coverage nodes. Could these be written in GFA, and how?

(@adamnovak @jeizenga the odgi version in libbdsg might be slow at handle iteration because it's always doing hash table lookups in the "hidden set".)

PR #130, opened 2020-06-19
ensure that the orientation of ranges is the same across encodings

subwaystation:
This brings in a change to the encodings or nucleotide ranges in the `odgi bin` JSON output. 
The ranges now follow this rule: 
Given range [a,b]:
1. a < b means forward orientation
2. a > b means reverse orientation.

### DIFF
--- a/src/algorithms/bin_path_info.cpp
+++ b/src/algorithms/bin_path_info.cpp
@@ -76,14 +76,20 @@ void bin_path_info(const PathHandleGraph& graph,
                                 (((nucleotide_count -  bins[curr_bin].ranges.back().first) > 1) && (bins[curr_bin].ranges.back().second == 0)) ||
                                 (is_rev != last_is_rev)){
 						    if (is_rev) {
-						        bins[curr_bin].ranges.push_back(std::make_pair(0, nucleotide_count));
-						    } else {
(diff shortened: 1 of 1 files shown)


subwaystation:
@dimatr Please take a look.

dimatr:
I'd suggest using the index trick
```
   std::get<is_rev ? 0 : 1>(foo) = 50; // first
   std::get<is_rev ? 1 : 0>(foo) = 50; // second
```

dimatr:
the lines 77-96 will boil down to smth like

```
                        int first_index = is_rev ? 0 : 1;
                        int second_index = is_rev ? 1 : 0;
                        (is_rev != last_is_rev)) {
                            std::pair p = std::make_pair(0, 0);
                            std::get<first_index>(p) = nucleotide_count;
			    bins[curr_bin].ranges.push_back(p);
			} else {
                            std::pair p = bins[curr_bin].ranges.back();
			    if (std::get<second_index>(p) == 0) {
			            std::get<second_index>(p) = std::get<first_index>(p);
			     }
			    std::get<first_index>(p) = nucleotide_count;
			}
```

subwaystation:
Thank you very much @dimatr for your feedback! My tests say it does what it should do. @ekg Only changes the JSON output, so your `.tsv` works as always ;)

PR #133, opened 2020-07-08
introduce weakly connected components in path guided 1D SGD sort

subwaystation:
Path guided 1D SGD can now deal with weakly connected components.

### DIFF
--- a/src/algorithms/path_sgd.cpp
+++ b/src/algorithms/path_sgd.cpp
@@ -698,25 +698,33 @@ namespace odgi {
                                  || (a.first == b.first
                                      && as_integer(a.second) < as_integer(b.second));
                       });
+#ifdef debug_components
             std::cerr << "node count: " << graph.get_node_count() << std::endl;
+#endif
             // refine order by weakly connected components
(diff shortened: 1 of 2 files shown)


subwaystation:
The current implementation introduces some memory overhead. I already have an idea of how to get rid of `std::vector<std::vector<handle_t>> weak_components_order` and directly inject the layout handle at its correct position in the final `order` vector. I will change `weak_components_map` so that it will hold the length of the component instead of its identifier. With this information, I can calculate where to insert the handle.

subwaystation:
@ekg If I understood correctly, I should also give `odgi sort` the option to output the components and their node identifiers. I would add another argument and emit all node identifiers for each component per line.

subwaystation:
> @ekg If I understood correctly, I should also give `odgi sort` the option to output the components and their node identifiers. I would add another argument and emit all node identifiers for each component per line.

I will just output the start index of each component into a file.  Very straightforward.

subwaystation:
Still work in progress. The final functionality is there and works for my tests.
I still want to replace the `weakly_connected_components` function with one that directly emits the HashMap I need at some point.

PR #135, opened 2020-07-12
pruning the graph without dropping paths

ekg:
This PR will work towards graph pruning without dropping paths. The first step is to allow removal of nodes without coverage (as this won't disrupt paths). The second is to split up paths where nodes have been removed, but not remove them from the graph.


ekg:
@subwaystation we talked about this some. It's not the best solution, but I think we should go for actually breaking the paths up. It's semantically coherent, if lossy.

subwaystation:
Yes, let's try it this way. You said there already exists a naming schema somewhere in `vg` about how to name broke paths?

PR #139, opened 2020-08-03
update libbf

ekg:
Point at upstream, thanks @JervenBolleman for the gcc 10.1 fix. @subwaystation also ran into the same thing.


subwaystation:
Works :)

PR #143, opened 2020-08-07
Path sgd

subwaystation:
Opening this up.


subwaystation:
@AndreaGuarracino This does not seem to work well.

DRB1-3123 fa gz gfa og Y I og

Putting the code into its own branch and then I will revert everything.

PR #145, opened 2020-08-07
Ranked handle graph

subwaystation:
Implement the ranked handle graph interface in odgi.


AndreaGuarracino:
Very nice, this will also improve the code readability, thank you!

PR #156, opened 2020-09-02
odgi chop preserves nodes order

AndreaGuarracino:


subwaystation:
This looks good to me @AndreaGuarracino. Chop-chop!

PR #163, opened 2020-09-07
odgi unchop preserves nodes order

AndreaGuarracino:


AndreaGuarracino:
Unchopping a chopped yeast pangenome (with max node length equals to 1), the node-order-preserving odgi unchop is only slightly slower than the not-order-preserving current master implementation.

```time odgi unchop -i Sc+Sp.pan.fpal10k.chopped1.og -o Sc+Sp.pan.fpal10k.chopped1.unchopped.og```

|time| master | order-preserving implementation |
| --- | --- | --- |
|real	| 2m27,693s | 2m28,514s |

but the outputs are very different

- on the left: the chopped yeast pangenome
- in the middle: the unchopped yeast pangenome with the order-preserving implementation
- on the right: the unchopped yeast pangenome with the master implementation (not order-preserving)

image (1)

**Comment by ekg** [comment]
Master's implementation is interesting.

Is there a smaller test case you can use to better understand the differences?

**Comment by AndreaGuarracino** [comment]
In this little example, only the 9-th and the 15-th nodes are chopped.

image

```
odgi build -g t.gfa -o - | odgi chop -i - -o - -c 9 -d | odgi viz -i - -o t.chop9.png
odgi build -g t.gfa -o - | odgi chop -i - -o - -c 9 -d | odgi unchop -i - -o - | odgi viz -i - -o t.chop9.unchop.master.png
odgi build -g t.gfa -o - | odgi chop -i - -o - -c 9 -d | /branch/odgi unchop -i - -o - | odgi viz -i - -o t.chop9.unchop.branch.png
```

image

On the left, the chopped visualization, in the middle the unchopped visualization using the master, on the right the unchopped visualization using the node-order-preserving implementation.

The difference is that with the master, the unchopped node takes each time the first 'free space'. The longest-node group is the first node group to be unchopped, and the new node (the old 9-th node in the original graph) goes at the end. The second node group to be unchopped (the old 15-th node in the original graph) becomes a new node that takes the place of the old longest-node group.

AndreaGuarracino:
Instead, the current branch outs each new node in place of the left-most node in the corresponding unchopped node group.

PR #164, opened 2020-09-08
improving path guided SGD

ekg:
Various hacks.

1. use xoshiro+ as a PRNG (we generate lots of random numbers)
2. use a "fast pow" to speed up our zipf distribution sampling

This is good for a ~40% speedup.


subwaystation:
Let me test this again.

subwaystation:
Works on my side :)

PR #169, opened 2020-09-17
odgi cover - nodes coverage initialization

AndreaGuarracino:
With this update, odgi cover initializes the nodes coverage considering the paths already embedded in the graph. Such paths can also be ignored during this initialization.


AndreaGuarracino:
This also allows you not to waste time if the nodes already have the required coverage (1 in the following example).

` time odgi cover -i yeast-pangenome/Sc+Sp.pan.fpal10k.og -c 1 -o a.og -t 16 -d`

> There will be generated paths until the minimum node coverage is 1, or until the maximum number of allowed generated paths is reached (23560690).
> SimpleCoverage: processing component 1 / 1, which has 4712138 node(s) and it is cyclic.
> Minimum node coverage reached after generating 0 paths.
> Processed: 1
> real	0m12,732s
> user	0m13,370s
> sys	0m1,135s

PR #170, opened 2020-09-21
Explore max eta

subwaystation:
This PR should explore, if we can have an ~30% faster convergence by fiddling around with the maximum learning rate.


subwaystation:
As a rather radical change, I set the default maximum learning rate to `num_path_steps` instead of `num_path_steps` <sup>2</sup>. The aim was to achieve ~the same metrics as with default settings, but with 20 iterations and a minimum learning rate of `0.05`. To account for our stochastic variability, I executed 5 runs for each data set evaluated.

### Results from the `DRB1-3123` data:

image
As expected, we are ~28% faster.
image
We are ~1% worse.
image
We are even 1% better.

subwaystation:
### Result from the `yeast` data:

image
I don't know why, but the performance on PG1 is very unstable.... we should not use it for benchmarks! We are ~23% faster....
image
We are ~34% worse.
image
We are 30% worse.

ekg:
Seems like we're getting into the realm of "no free lunch"

Hope not.

subwaystation:
### Results from `1001G_Chr1` data. This time we set `-v 1`!
image
Still the performance on PG1 is super unstable. We are ~21% faster.
image
We are ~17% better.
image
We are ~18% better. @ekg 

Hmm. I think it is worth to try out `-v 1` with the yeast data set.
But as @AndreaGuarracino mentioned to me I might be fishing blindly here.

subwaystation:
@ekg I don't want to give up our free lunch, yet.

AndreaGuarracino:
My feeling is that with a blind learning rate, there will never be a good combination of default values. We could find the best one for `few` datasets, but not for the `all-few` others. I don't think (and hope) we are wasting ~30% of the time for all the datasets.

I never tried with the `1001G_Chr1` data, but with a so low learning rate, the impression is that the SGD is working, but not so much. Maybe it can be useful for us to see also the max `delta`s with when `-v 1`.

subwaystation:
### Results from `yeast` with `-v 1`:

image
~22% faster.
image
~37% worse.
image
~31% worse.

subwaystation:
To summarize this short experiment:

If the parameters are set in a favorable way for the data set, we should be able to reduce the runtime by 45-50 percent without making the sort worse (see 1001G_Chr1 results). However, the other results indicate, that we have to find a generally applicable formula in order to make this possible for all data sets.

PR #173, opened 2020-09-28
odgi viz/bin improvements and porting binned mode functionalities in default mode

AndreaGuarracino:
In binned mode, the bin links are gray, not black

`odgi viz -i x.gfa -o x.b.png -b`

x b

`odgi viz -i x.gfa -o x.bm.png -bm`

x bm

`odgi viz -i x.gfa -o x.bz.png -bz`
x bz

In this way, we can quickly distinguish the type of links and the visualization mode which generated the images.

In addition, in default mode, forward/reverse nodes can be colored black/red. This can be useful when, for example, a path in the dataset is totally reversed, but contiguous:

`odgi viz -i x.gfa -o x.du.png -du`

x du

`odgi viz -i x.gfa -o x.bz.png -bz`

x bz

`odgi viz -i x.gfa -o x.z.png -z`

x z


AndreaGuarracino:
Added the percentage of gap links removed in odgi bin and odgi viz to have a quicker look at what we have (for example, to judge how good a sorting is).

`odgi viz -i Sc+Sp.pan.fpal10k.Y.G50.og -o Sc+Sp.pan.fpal10k.Y.G50.png -z -x 10000 -bg`

```
Binned mode
bin width: 1983.09
image width: 10000
Gap links removed: 24.44%, that is 32594 gap links (192 path start links + 192 path end links + 32210 inner gap links) of 133362 total links
```

PR #175, opened 2020-10-07
added progress meter to p_sgd

subwaystation:


ekg:
Mind adding it to the 2D case? Sorry we are duplicating code still.

subwaystation:
Sure. I forgot about that.

Issue #129, opened 2020-06-16
path guided linear 1D SGD sort - edge cases


subwaystation:
The following graphs break the sorting algorithm because each path walks through exactly one node. We may have to address this at some point.
[DRB5-3127.gfa.txt](https://github.com/vgteam/odgi/files/4785270/DRB5-3127.gfa.txt)
[V-352962.gfa.txt](https://github.com/vgteam/odgi/files/4785271/V-352962.gfa.txt)

subwaystation:
@ekg discovered this in #127 .

josiahseaman:
How about a special case statement that tries to put the first node of a
path at the beginning when it does not share any nodes with the list of
sorted nodes?

PR #181, opened 2020-10-16
path-guided sorting can work only with paths with at least two steps

AndreaGuarracino:
To avoid error like

```odgi: /home/tools/odgi/build/sdsl-lite-prefix/src/sdsl-lite-build/include/sdsl/int_vector.hpp:1405: sdsl::int_vector<<anonymous> >::const_reference sdsl::int_vector<<anonymous> >::operator[](const size_type&) const [with unsigned char t_width = 0; sdsl::int_vector<<anonymous> >::const_reference = const long unsigned int; sdsl::int_vector<<anonymous> >::size_type = long unsigned int]: Assertion `idx < this->size()' failed.```

when there are paths of just 1 node.

### DIFF
--- a/src/algorithms/path_sgd.cpp
+++ b/src/algorithms/path_sgd.cpp
@@ -90,6 +90,7 @@ namespace odgi {
 
                 if (path_index.get_path_step_count(path) > 1){
                     at_least_one_path_with_more_than_one_step = true;
+                    break;
                 }
             }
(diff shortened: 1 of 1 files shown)


AndreaGuarracino:
Already known edge case... #127 #129

AndreaGuarracino:
Already known edge case: #127. It closes #129.

PR #182, opened 2020-10-21
path-guided layout can work only with paths with at least two steps 

AndreaGuarracino:
To avoid in 2D errors as in #129.


AndreaGuarracino:
The error occurred when trying to calculate the layout (`odgy layout`) of smoothed graphs with consensus sequences of just one node.

PR #195, opened 2020-11-30
Fastify

ekg:
This breaks the odgi format significantly.

Previously, we were storing path steps as deltas. This was efficient when the graph is sorted. Packing and unpacking these relativistic encodings is not trivial, so in this branch the idea is to do away with these operations, replacing them with per-node alphabet encoder that project the local neighborhood of nodes reached by edges and path steps into a small alphabet.

Path steps now record on special bits if they are the start and/or end of the given path.

The encoding scheme is not yet implemented, but the backing storage has been ported.


ekg:
This is a significant change. I'll tag a release for it.

```
erik@beyyo [03:36:07 PM] [~/odgi] [fastify]
-> % git diff master | wc -c
167260
```

PR #199, opened 2020-12-11
also to unlock the handles we need to lock the lockers

AndreaGuarracino:
This fixes a problem in the handles locking that occurs more frequently when the number of threads is high.


subwaystation:
I just checked this out in `smoothxg` and it fixes my problem. Great job!

PR #227, opened 2021-03-09
HaploBlocker output for odgi bin

subwaystation:
This will work on an odgi bin output specifically curated for HaploBlocker https://github.com/tpook92/HaploBlocker.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -76,6 +76,3 @@
 [submodule "deps/atomicbitvector"]
 	path = deps/atomicbitvector
 	url = https://github.com/ekg/atomicbitvector.git
-[submodule "deps/gzstream"]
-	path = deps/gzstream
-	url = https://github.com/kanedo/gzstream.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -76,3 +76,6 @@
 [submodule "deps/atomicbitvector"]
 	path = deps/atomicbitvector
 	url = https://github.com/ekg/atomicbitvector.git
+[submodule "deps/gzstream"]
+	path = deps/gzstream
+	url = https://github.com/kanedo/gzstream.git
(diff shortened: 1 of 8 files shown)


subwaystation:
@tpook92 I did some testing on the yeast pangenome graph of length 14778003 with 112 paths. The construction time for such a graph usually is around 3 minutes on my laptop.

In the following an overview of my `odgi bin -b` runs:

|      | time m:ss  | RAM kbytes  | file size MB  |   |
|------|---|---|---|---|
| w1   | 1:16.88  |  541420 | 2900  |   |
| w10  |  0:09.84 | 267928  | 289  |   |
| w100 |  0:03.18 | 251832  | 29  |   |
| w1_pigz | 1:17.72  | 541464  | 31  |   |

If the generation of the binned output takes ~1/3 of the construction time, that's a little bit slow. The major reason is that the TSV format we discussed is very verbose. I often observe bins where only 2 paths have coverage. But for all other paths I have to write `0`. This would not be the case using e.g. JSON.

So two points to discuss:
1. Is this file format really the way to go forward? A JSON format only emitting each present bin of each path would be much less verbose. But maybe harder for you to handle in R?
2. Maybe it is sufficient to go for a bin size of e.g. 10. But ensuring that we have at least 0.8 coverage for each bin and path. Else I would not report it in the TSV.

subwaystation:
Alright @tpook92. Another round :)

This time with the line-wise JSON format as discussed.

|      | time m:ss  | RAM kbytes  | file size MB  |
|------|---|---|---|
| w1   | 0:27.76  | 536628  | 979  |
| w10  |  0:04.75 | 278272  | 92  |
| w100 |  0:02.69 | 253036  | 8  |
| w1_pigz | 0:29.02  | 250599  | 186  |

We see great improvements in runtime and file sizes. Yihaa!
Let's try out some human data next.

PR #230, opened 2021-03-16
Meanderless

ekg:
Fixes a major issue with path-guided graph sorting.

The radius of the zipf function we choose causes curls in the layout. These appear to occur for 1D and 2D, but they are easier to understand in 2D.

This can be seen in this plot of a yeast pangenome:

Screenshot from 2021-03-15 12-36-25

The curving can be mitigated by sampling the pair of steps used to drive any update to the layout model completely randomly half of the time. The other half, we continue to use zipf sampling, thus tending to sample nodes that are close together.

Screenshot from 2021-03-16 19-23-38

### DIFF
--- a/src/algorithms/path_sgd.cpp
+++ b/src/algorithms/path_sgd.cpp
@@ -230,32 +230,40 @@ namespace odgi {
                                         continue;
                                     }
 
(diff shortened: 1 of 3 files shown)


ekg:
Screenshot from 2021-03-16 21-44-53

after integrating this in smoothxg / pggb

PR #237, opened 2021-03-18
odgi extract preserves path order

AndreaGuarracino:
This PR will slow down a bit the extraction, but with the difference that the path order in the subgraph will be preserved and deterministic.


AndreaGuarracino:
I withdraw what I wrote: refactoring a bit of the code, the overhead is just about 1 or 2 seconds to extract the centromeres of the human chr2 pangenome, build with 90 haplotypes plus two references.

PR #262, opened 2021-04-27
Towards an index free PG-SGD

subwaystation:
We want to have an index free 1D and 2D PG-SGD. 
 - As an intermediate solution, we store all `step_handle_t` in two `uint64_t` vectors. Later we will find a way to directly sample a `step_handle_t`. Maybe a `ska::flat_hash_map` could be a solution.
 - We can retrieve the `handle_t` from a `step_handle` using the `odgi` graph itself.
 - We can retrieve the `path_handle_t` from a `step_handle` using the `odgi` graph itself.
 - The first `step_handle_t` is sampled uniformly.
 - The second `step_handle_t` is sampled using the zipfian distribution to directly sample a nucleotide distance from the first`step_handle_t`. Once we know the distance we can traverse the graph using `get_next_step(step_handle_t step_handle)` for all possible ranges of this instance. Presumably left and right. Each update is recorded.
 
 The new algorithm in short:
 
1. Get a random step, uniformly drawn.
2. Get a random nucleotide distance, Zipfian drawn.
3. Walk the path until we have reached the nucleotide distance leading to the second step.
4. Run an update of both nodes by the two given steps.
5. Keep incrementing both steps by the sampled nucleotide distance by walking the graph until we get to the end of the path.
6. Same as in 5, but increment to the other direction.

This might change how many term updates we need to do.

### DIFF
--- a/src/algorithms/path_sgd_layout.cpp
+++ b/src/algorithms/path_sgd_layout.cpp
@@ -49,11 +49,6 @@ namespace odgi {
             snapshot_progress[0].store(true);
             // seed them with the graph order
             uint64_t len = 0;
-            // the longest path length measured in nucleotides
-            size_t longest_path_in_nucleotides = 0;
-            // the total path length in nucleotides
-            size_t total_path_len_in_nucleotides = 0;
-
             bool at_least_one_path_with_more_than_one_step = false;
 
             for (auto &path : path_sgd_use_paths) {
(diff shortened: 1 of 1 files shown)


ekg:
We could just sample a path (by length) then run out the sampled distance, then step along the path until we hit the end.

subwaystation:
We might want to replace the current index with the step index. Will do another PR for this.

PR #287, opened 2021-06-24
untangle loops into paired BED

ekg:
`odgi untangle` lets us linearize looping structures in the graph into BEDPE format.

A greedy and complete segmentation algorithm first breaks each user-specified query and reference path into segments where the path does not loop onto itself. We index the reference segments to let us quickly find the set of reference segments at a given node. Then, for each query segment, we find the reference segment that has the highest jaccard calculated in terms of shared sequence in the graph.

In effect, we can express a given query path as a patchwork of reference segments. Where graphs are simple and linear, most mappings are 1:1. But, in looping regions (segdups and VNTRs), we can describe a given sample as a rearrangement of a set of reference haplotypes.

This is a work in progress. Current todo:

- ~~The command line inputs need updating to allow for reference and query lists, or prefix-based selection of paths.~~
- ~~Relative orientation should be written in the output.~~
- ~~A recursive call used to implement the segmentation should be replaced with a heap-allocated stack or queue (so we can scale to centromere-level looping).~~
- We should segment in two passes, the first to obtain all self-optimal segmentations, which we then use to seed a second round of segmentation which will ensure that even non-looping paths get segmented where other paths loop.
- Independent processing steps should be run in parallel (segmentation, reference segment indexing, and querying).

Example on CYP2D6:

image

image

`-> % odgi untangle -i cyp2d6.og -q CYP2D6_all__HG00438__mat__ID.1 -r CYP2D6_all__CHM13__pri__ID.1 -m 32`

```
CYP2D6_all__HG00438__mat__ID.1  0       55856   CYP2D6_all__CHM13__pri__ID.1    0       55774   0.993001
CYP2D6_all__HG00438__mat__ID.1  55856   63281   CYP2D6_all__CHM13__pri__ID.1    55774   63199   0.99115
CYP2D6_all__HG00438__mat__ID.1  63281   68249   CYP2D6_all__CHM13__pri__ID.1    63199   68168   0.991383
CYP2D6_all__HG00438__mat__ID.1  68249   76991   CYP2D6_all__CHM13__pri__ID.1    68168   76869   0.90405
CYP2D6_all__HG00438__mat__ID.1  76991   81962   CYP2D6_all__CHM13__pri__ID.1    63199   68168   0.990787
CYP2D6_all__HG00438__mat__ID.1  81962   90661   CYP2D6_all__CHM13__pri__ID.1    68168   76869   0.991302
CYP2D6_all__HG00438__mat__ID.1  90661   187666  CYP2D6_all__CHM13__pri__ID.1    76869   173861  0.99569
```


subwaystation:
:fire:

ekg:
This is working on everything I've tested it against. Note that when run with --threads > 1 the output BED will not be sorted.

Issue #282, opened 2021-06-21
odgi build error with GFA graph


brettChapman:
Hi

I'm getting an error with one of my graphs. I'm trying to build from GFA and am getting the following error:

```
odgi build -t 32 -P -g barley_pangenome_graph_2H.gfa -o barley_pangenome_graph_2H.gfa.og
[odgi::gfa_to_handle] building nodes: 73.86% @ 1.99e+06/s elapsed: 00:00:00:53 remain: 00:00:00:19odgi: /smoothxg/deps/odgi/src/odgi.cpp:507: virtual handlegraph::handle_t odgi::graph_t::create_handle(const string&, const nid_t&): Assertion `!has_node(id)' failed.
srun: error: node-9: task 0: Aborted (core dumped)
```

The original graph has consensus paths in it, so I removed them. I then converted from VG to GFA.

Is there a way I can process the graph either from the VG or GFA to clean it up and remove what appears to be a node ID causing a particular error. I don't think its a memory issue as I checked for any killed processes on the compute node, and there were none. I also have 126GB RAM on the node, and the GFA file is only 25GB is size.

subwaystation:
Hi @brettChapman 

The error message indicates that you have two times the same node identifier in your GFA. One way to verify this is to compare the total number of node identifers with the unique identifiers:

``` bash
grep "^S" barley_pangenome_graph_2H.gfa | cut -f 2 | wc -l
grep "^S" barley_pangenome_graph_2H.gfa | cut -f 2 | uniq | wc -l
```
If both numbers are the same, then this needs further investigation on the `odgi` side. 

Or you can also validate you GFA with `vg validate`. I don't think a GFAv1 is allowed to have duplicated node identifiers. So the file you have could have become malformed at some point.
In `vg` there is the `ids` subcommand, but I don't think it will fix your problem.

How did you built your GFA?

brettChapman:
Hi @subwaystation

I first generated the GFA through PGGB. I then removed consensus paths and generated VG graphs.

I then ran vg ids to get the node id unique across all graphs:
```
vg ids -j $(for i in $(seq 1 7); do echo barley_pangenome_graph_${i}H.vg; done)
```

Now with unique node IDs, I indexed the graphs to generate one large index, so I could run vg gbwt.

```
vg index -x barley_pangenome_graph.xg $(for i in $(seq 1 7); do echo barley_pangenome_graph_${i}H.vg; done)
```
```
[...]

PR #288, opened 2021-06-28
replacing dset64-gccAtomic.hpp with dset64.hpp

subwaystation:
We want to be able to build `odgi` without the `-mcx16` GCC option. Therefore, we are replacing the `dset64-gccAtomic.hpp` with `set64.hpp`. This is also an indirect attempt to solve #282.


subwaystation:
ADIOS `dset64-gccAtomic.hpp` !

PR #289, opened 2021-06-29
odgi viz color by coverage

ekg:
The color scheme can be chosen from anything in colorbrewer2: https://colorbrewer2.org


ekg:
C4 with different colorings:

RdPu:5

image

Spectral:5

image

PR #295, opened 2021-07-02
scale untangle to whole chromosomes

ekg:


ekg:
The problem is in the positional index (hash table based). This should use the XP, which could use a little bit of work to get it to construct in parallel.

PR #301, opened 2021-07-20
replace all optimize checks with 'is_optimized()' so that we have a stringend check, if a graph is optimized

subwaystation:
We need to monitor carefully, if a graph is optimized or not, or some of our subcommands break. A graph is seen as optimized if its `min_node_id == 1` and if it its `max_node_id == node_count`. This PR fixes this in all subcommands.


subwaystation:
@brettchapman This might be of particular interest to you! Please ensure that when you build your graph to also use `-O, --optimize`. I tried `odgi viz` with one of your graphs, not optimized, and it ran forever.

AndreaGuarracino:
`odgi viz` doesn't need such a strong requirement. Probably, also `odgi sort` and `odgi layout`.

subwaystation:
> `odgi viz` doesn't need such a strong requirement. It can be extended. Probably, also `odgi sort` and `odgi layout`.

How? Could you please push something? Thanks!

AndreaGuarracino:
The optimization requirement is too strong for the majority of the commands involved in this PR. Have a look here https://github.com/pangenome/odgi/pull/302.

odgi `viz`, `depth`, `degree`, and `stats` can work with strictly optimized graphs.\
odgi `extract` doesn't need an optimized graph, just that the IDs are consecutive and compact, so it was already fine.

The optimization requirement could be left for odgi `sort`, `layout`, and `pathindex`, but it could be avoided also there with a bit more work.

subwaystation:
This is ready to go. At some point in time we should validate all our subcommands and find out, if some of them require an optimized graph, too. To guard them against non-optimized graphs.

PR #302, opened 2021-07-20
odgi `viz`, `depth`, `degree`, and `stats` can work with not optimized, but compacted graphs

AndreaGuarracino:
This tackles the problem encountered here https://github.com/pangenome/odgi/pull/301

This could be extended also for odgi `sort`, `layout`, and `pathindex`, but it would require changes in the XP index.

### DIFF
--- a/src/subcommand/stats_main.cpp
+++ b/src/subcommand/stats_main.cpp
@@ -142,7 +142,15 @@ int main_stats(int argc, char** argv) {
         }
     }
 
-    if (yaml) {
+	const uint64_t shift = number_bool_packing::unpack_number(graph.get_handle(graph.min_node_id()));
+
+    if (args::get(mean_links_length) || args::get(sum_of_path_node_distances) || yaml) {
+		if (number_bool_packing::unpack_number(graph.get_handle(graph.max_node_id())) - shift >= graph.get_node_count()){
+			std::cerr << "[odgi::stats] error: the node IDs are not compacted. Please run 'odgi sort' using -O, --optimize to optimize the graph." << std::endl;
+			exit(1);
+		}
+    }
+	if (yaml) {
     	std::cout << "---" << std::endl;
     }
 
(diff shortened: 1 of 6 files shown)


src/subcommand/stats_main.cpp
```diff
@@ -141,7 +141,14 @@ int main_stats(int argc, char** argv) {
 			utils::handle_gfa_odgi_input(infile, "stats", args::get(progress), num_threads, graph);
         }
     }
-    ///graph.display();
+
+    /// we need to check in advance if the graph is compacted
```
AndreaGuarracino, line 145: The compacted IDs are needed only where we use the `position_map` vector, that is when `mean_links_length` and/or `sum_of_path_node_distances` are computed.

subwaystation (approved):
Alright, I think this is ready to go. We can merge this, which will override some changes I made in https://github.com/pangenome/odgi/pull/301/. 
I looked again at `sort`, `layout`, and `pathindex`. The code in `xp.cpp` can become fairly complex and interesting to test, if we would go the same way as in this PR.
So for these 3 subcommands, I would suggest to keep the hard optimize definition avoiding downstream bugs. 

What do you think @AndreaGuarracino ?

src/subcommand/stats_main.cpp
subwaystation: We need to have this check if we have `-y, --yaml`.

subwaystation: Let me push something.

subwaystation: Done.

subwaystation:
@AndreaGuarracino Please merge, if you are happy with it :+1:

AndreaGuarracino:
Prophecy: we will talk about this again, but for now the show must go on

subwaystation:
> Prophecy: we will talk about this again, but for now the show must go on

Afraid so. Fingers crossed!

PR #325, opened 2021-09-24
`odgi position` - add the path jaccard concept when translating path to path positions

subwaystation:
This adds the path jaccard concept, an alignment-free local sequence similarity threshold, to `odgi position`. It allows a more precise path position translation when the path we want to translate to has several candidate steps and we find the best one using the jaccard concept. Borrows functionality from `odgi tips` and therefore works equivalently.

- [x] Implement the path jaccard in `odgi position`.
- [x] Refactor and beautify code. Code used both in `odgi tips` and `odgi position` should be moved into `path_jaccard.*pp`.
- [x] Add binary tests. Only makes sense if we can make use of GitHub's workflow caching functionality. Write one `bash` file for one subcommand, in this case only `odgi position`.
- [x] Clarify why some default `odgi position` tests cases don't produce an output. I suspect the BFS.
- [x] Update the documentation of `odgi position`. 
- [x] If several candidate steps have the same jaccard, we need to chose one of these candidates deterministically. The idea is the take a look at all the ranks of the steps and take the ranks/2 one. If the result is not dividable by two, we take the lower value.


subwaystation:
@ekg
I have the following graph:
overlap gfa
Here as GFA [overlap.gfa.txt](https://github.com/pangenome/odgi/files/7226068/overlap.gfa.txt).
For the following `odgi position` invocations I would expect some output, but I don't get any!
``` bash
odgi position -i overlap.gfa -r target -p query3,0
odgi position -i overlap.gfa -r target -p query3,1
odgi position -i overlap.gfa -r target -p query3,2
odgi position -i overlap.gfa -r target -p query3,3

odgi_ position -i overlap.gfa -r target -p query1,0
odgi_ position -i overlap.gfa -r target -p query1,1
```
I suspect it is the BFS's fault? Because we invoke it with the head not of the `target` it is somehow not capable to take a look back?! Any spontaneous ideas? Thanks :)

subwaystation:
If I understood the BFS correctly, it is not going bidirectional, although it should.

subwaystation:
It knows it should go bidirectional:

```
odgi_ position -i overlap.gfa -r target -p query3,0
bidirectional: 0
bidirectional: 1
```

subwaystation:
Not flipping the handle in https://github.com/pangenome/odgi/blob/8900d23e02bcc60cd73f39f3ddabf81c4ecc8ceb/src/subcommand/position_main.cpp#L522 gives me the correct result for the above, e.g.
```
odgi_ position -i overlap.gfa -r target -p query3,0
#source.path.pos	target.path.pos	dist.to.ref	strand.vs.ref
query3,0,+	target,0,+	3	-
```
But all other queries which worked before are now wrong. Hmm.

subwaystation:
Alright, I found the problem. The BFS couldn't go bidirectional, because the start handle was still in the list of seen nodes. Also, I had to adjust the positions when we go bidirectional. Now all tests work, especially:

```
[heumos@savasana test]$ odgi_ position -i overlap.gfa -r target -p query3,0
#source.path.pos	target.path.pos	dist.to.ref	strand.vs.ref
query3,0,+	target,0,+	3	+
[heumos@savasana test]$ odgi_ position -i overlap.gfa -r target -p query3,1
#source.path.pos	target.path.pos	dist.to.ref	strand.vs.ref
query3,1,+	target,1,+	3	+
[heumos@savasana test]$ odgi_ position -i overlap.gfa -r target -p query3,2
#source.path.pos	target.path.pos	dist.to.ref	strand.vs.ref
query3,2,+	target,2,+	3	+
[heumos@savasana test]$ odgi_ position -i overlap.gfa -r target -p query3,3
#source.path.pos	target.path.pos	dist.to.ref	strand.vs.ref
query3,3,+	target,1,+	1	+
[heumos@savasana test]$ odgi_ position -i overlap.gfa -r target -p query3,4
#source.path.pos	target.path.pos	dist.to.ref	strand.vs.ref
query3,4,+	target,1,+	0	+
```

PR #329, opened 2021-09-30
`odgi extract`: be more specific about path lacing options

subwaystation:
I just ran into a problem where I could produce a subgraph with the paths laced back together, because the documentation of the option `-E, --full-range` was insufficient. Fixed with this PR.

### DIFF
--- a/docs/rst/commands/odgi_extract.rst
+++ b/docs/rst/commands/odgi_extract.rst
@@ -68,7 +68,7 @@ Extract Options
 
 | **-E, --full-range**
 | Collects all nodes in the sorted order of the graph in the min and max
-  position touched by the given path ranges. Be careful to use it with
+  position touched by the given path ranges. This ensures that all the paths of the subgraph are not split by node, but that the nodes are laced together again. Comparable to **-R, --lace-paths=FILE**, but specifically for all paths in the resulting subgraph. Be careful to use it with
   very complex graphs.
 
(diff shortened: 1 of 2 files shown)


AndreaGuarracino:
`This ensures that all the paths of the subgraph are not split by node, but that the nodes are laced together again. Comparable to **-R, --lace-paths=FILE**, but specifically for all paths in the resulting subgraph.`

This is not how `-E` works. It just takes all nodes with IDs between the minID and the maxID. This can work well with well-sorted graphs.

subwaystation:
```
        -E, --full-range                  Collects all nodes in the sorted order
                                          of the graph in the min and max
                                          positions touched by the given path
                                          ranges. Be careful to use it with very
                                          complex graphs.
```                  
But it ensures that the paths are laced back again. At least from my understanding.

subwaystation:
> `This ensures that all the paths of the subgraph are not split by node, but that the nodes are laced together again. Comparable to **-R, --lace-paths=FILE**, but specifically for all paths in the resulting subgraph.`
> 
> This is not how `-E` works. It just takes all nodes with IDs between the minID and the maxID. This can work well with well-sorted graphs.

So for each path, it takes all the nodes for that path in the subgraph and ensures that the path traverses these nodes from minID to maxID of these nodes. Which is what I would call lacing. Am I correct here?

PR #341, opened 2021-10-29
`odgi untangle` power up - also include the segments at the tips of each query path

subwaystation:
Trying to teach `odgi untangle` to also incorporate the tips of each query path when untangling them.

- [X] Implement prototype.
- [x] Test prototype on small scale data.
- [x] Test prototype on large scale data.


ekg:
This does seem to work on the yeast dataset.

PR #350, opened 2021-11-24
Minor fixes for building with Clang 13

sampsyo:
I tried building odgi with clang 13.0.1 on Ubuntu 18.04 today, and mostly things went pretty well. But I ran into a couple of very minor C++ wrinkles—hopefully neither of these is too surprising:

- The `d` suffix on double literals appears to be nonstandard and not supported by Clang.
- `progress.hpp` uses `std::stringstream`, which is from the `sstream` header that was not being included.

Let me know if any of this looks wrong; happy to investigate further if so!

### DIFF
--- a/src/algorithms/atomic_image.cpp
+++ b/src/algorithms/atomic_image.cpp
@@ -66,7 +66,7 @@ color_t layer(const color_t& a, const color_t& b, const double& f) {
     //std::cerr << "got a = " << a.hex << " " << (int)a.c.r << "," << (int)a.c.g << "," << (int)a.c.b << std::endl;
     //std::cerr << "got b = " << b.hex << " " << (int)b.c.r << "," << (int)b.c.g << "," << (int)b.c.b << std::endl;
     //std::cerr << "f = " << f << std::endl;
-	if (.0d < f) {
+	if (.0 < f) {
 		out.c = (RGB) { inrange((double)b.c.r - ((double)(255 * (1-f) - a.c.r))),
                         inrange((double)b.c.g - ((double)(255 * (1-f) - a.c.g))),
                         inrange((double)b.c.b - ((double)(255 * (1-f) - a.c.b))),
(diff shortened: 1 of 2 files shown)


subwaystation:
Hi @sampsyo,
cool that you are updating ODGI to be built via clang :)
Thanks for the PR.
Once @ekg reviewed your PRs to his repository, we can ship these with this PR.

As I don't know clang, what would I have to execute to build ODGI with it? Despite the clang package, would there be any other dependencies on Ubuntu?
Once this flies, I would like to add this alternative building approach to our Docs and add a new GitHub building workflow.

Your proposed changes here don't break the cmake build or subsequent tests, so that's good ;)

sampsyo:
Sure! In my case, I got a recent Clang from [LLVM's apt repository](https://apt.llvm.org). (I needed the packages `clang-13 libc++-13-dev libc++abi-13-dev libomp-13-dev`.) Then, after fixing a few things, it was as simple as telling CMake which compiler (and C++ stdlib) to use:

    CC=clang-13 CXX=clang++-13 CXXFLAGS=-stdlib=libc++ cmake -H. -Bbuild
    CC=clang-13 CXX=clang++-13 CXXFLAGS=-stdlib=libc++ cmake --build build -- -j

I didn't run into any other missing dependencies.

ekg:
It looks like we can still merge this. Does it make sense to you @subwaystation ?

subwaystation:
For building on Archlinux I had to execute:

```
CC=clang-13 CXX=clang++ CXXFLAGS=-stdlib=libc++ cmake -H. -Bbuild
CC=clang-13 CXX=clang++ CXXFLAGS=-stdlib=libc++ cmake --build build -- -j
```

Then I ran into 
```
In file included from /home/heumos/git/odgi/src/algorithms/sgd_layout.cpp:2:
/home/heumos/git/odgi/deps/sgd2/src/sgd2.hpp:96:14: error: no member named 'random_shuffle' in namespace 'std'
        std::random_shuffle(terms.begin(), terms.end());
        ~~~~~^
/home/heumos/git/odgi/deps/sgd2/src/sgd2.hpp:466:14: error: no member named 'random_shuffle' in namespace 'std'
        std::random_shuffle(terms.begin(), terms.end());
        ~~~~~^
```
as already pointed out by @sampsyo.
@ekg Could you please take a look and probably merge https://github.com/ekg/dirtyzipf/pull/1 and https://github.com/ekg/sgd2/pull/2? Thanks!

PR #351, opened 2021-11-25
`odgi extract` powerup: extract pangenomic ranges

AndreaGuarracino:
This allows users to extract subgraphs by specifying pangenomic ranges, that is nucleotide ranges that refer to the pangenome’s sequence (i.e., the sequence obtained arranging all the graph’s node from left to right). This can be useful in specific cases.

**Example: extract sequences that do not align to the reference**
Given this rat chr1 pangenome as an example:

image

we can see sequences on the left/right that do not align with the reference (`REF`). A simple way to extract that region is to specify a pangenomic range from 0 to *something*.

`odgi extract -i chr1_mratnor1.pan+ref.fa.gz.f7b23b2.4030258.ab53614.smooth.og -q 0-145000000 -t 48 -P -o - | odgi viz -i - -o x.png -M sample.names`
image

In this case, we are extracting also a little piece of the reference. Part of these sequences presents uncalled bases:

image

Extracting and visualizing is a fast-enough workflow for searching a pangenomic range of interest, at least for not huge pangenomes, but we still need better ways to find such a range more easily and interactively.

### DIFF
--- a/docs/rst/commands/odgi_extract.rst
+++ b/docs/rst/commands/odgi_extract.rst
@@ -63,8 +63,11 @@ Extract Options
 | Find the node(s) in the specified path range TARGET=path[:pos1[-pos2]]
   (0-based coordinates).
 
-| **-r, --bed-file**\ =\ *FILE*
+| **-b, --bed-file**\ =\ *FILE*
 | Find the node(s) in the path range(s) specified in the given BED FILE
+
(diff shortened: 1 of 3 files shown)


ekg:
We should add a way to draw vertical lines on the odgi viz plot that
correspond to either pangneomic positions or path ranges. Or even adding
them as rows in the path matrix would help to make the selection.

PR #353, opened 2021-12-07
a sampled step index

subwaystation:
The current step index implementation is very heavy when it comes to RAM usage. This PR attempts to solve that by only indexing every node with node identifier fitting  `mod(node_id, step-index-sample-rate) == 0` in the graph. 
From a given step, we can find its position by walking backwards until a node fitting our sampling criteria is found. We can retrieve this position easily, adding up the walked distance to retrieve the actual position of the step. 
Effectively, the sample rate is only allowed to be a number by the power of 2, because we can use bit shift operations to calculate the modulo in O(1)! (https://www.geeksforgeeks.org/compute-modulus-division-by-a-power-of-2-number/).

- [x] Basic implementation.
- [x] Can we add a pointer to a graph_t in the step_index_t so that we don't need a graph_t at hand for the `get_position(step, graph)` function? --> We will need the graph at all times anyhow, so it won't matter.
- [x] We need to figure out which value for the sampling rate delivers the best compromise of performance and memory usage. The current default in `odgi tips` and `odgi untangle` is 8.
- [X] Add unittests.
- [x] Add the length of each path to the index. This is a requirement to replace the XG in `smoothxg`. Might be tricky to serialize to disk, then. Not sure here.
- [x] Find a way to serialize and deserialize a step index to disk.
- [x] Update unittests.
- [x] Update the stepindex, tips, untangle, and position interfaces.
- [x] Binary tests for tips, untangle.
- [x] Update the Docs.

### DIFF
--- a/src/algorithms/stepindex.cpp
+++ b/src/algorithms/stepindex.cpp
@@ -7,7 +7,15 @@ namespace algorithms {
 step_index_t::step_index_t(const PathHandleGraph& graph,
                            const std::vector<path_handle_t>& paths,
                            const uint64_t& nthreads,
-                           const bool progress) {
+                           const bool progress,
+						   const uint64_t sample_rate) {
+	// if the sample rate is not dividable by 2, the algorithm might be slow
+	if (sample_rate % 2 != 0) {
+		if (progress) {
+			std::cerr << "[odgi::algorithms::stepindex] Validating sample rate: The given sample rate of " << sample_rate << " is not dividable by 2. This will slow down the building and usage of the step index.0" << std::endl;
+		}
+	}
+	this->sample_rate = sample_rate;
     // iterate through the paths, recording steps in the structure we'll use to build the mphf
     std::vector<step_handle_t> steps;
 	std::unique_ptr<algorithms::progress_meter::ProgressMeter> collecting_steps_progress_meter;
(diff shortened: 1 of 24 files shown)


subwaystation:
After discussing the evaluation https://docs.google.com/presentation/d/1a8bOnulta6fYnQ2DFmdzt4es2vaRGmgIxO3kCe-HXR8/edit#slide=id.p with @ekg and @AndreaGuarracino we decided to go for a default sampling rate of 8. However, if we should blow out of memory on insanely large graphs, a sampling rate of 16 would still be highly executable.

src/algorithms/stepindex.cpp
```diff
@@ -46,21 +56,51 @@ step_index_t::step_index_t(const PathHandleGraph& graph,
...
     }
 	if (progress) {
 		building_progress_meter->finish();
 	}
+	// TODO add path lengths here
 }
 
-const uint64_t& step_index_t::get_position(const step_handle_t& step) const {
-    return pos[step_mphf->lookup(step)];
+const uint64_t step_index_t::get_position(const step_handle_t& step, const PathHandleGraph& graph) const {
```
subwaystation, line 79: @ekg With this function, we could print out all reference positions, if we iterate through all steps of a given reference path.

src/algorithms/stepindex.hpp
```diff
@@ -1,12 +1,14 @@
 #pragma once
 
+// FIXME including this breaks the compilation
+#include <sdsl/enc_vector.hpp>
```
subwaystation, line 4: @ekg This line breaks the compilation.

subwaystation: Got it working.

PR #365, opened 2022-02-15
Improved guix support with guix.scm

pjotrp:
Updated README and added guix.scm. It can build odgi on top of a recursive git checkout. 

This is the starting point for getting dependencies sorted in Guix itself.


subwaystation:
:octopus:

PR #375, opened 2022-03-15
odgi-ffi: started on C-API

pjotrp:


pjotrp:
Made a separate PR with a single patch.

PR #392, opened 2022-04-30
Inject BED range annotations as paths

ekg:
This allows the input of BED ranges as paths inside the graph. There are many potential applications of embedding path-based annotations. Embedded annotation paths can be used to pick up part of the graph in `odgi extract`, or as targets for liftover in `odgi untangle`.

In particular, this feature is useful to generate gene arrow diagrams using [gggenes](https://github.com/wilkox/gggenes) and the output of `odgi untangle`. I'll work up a tutorial to merge in this PR.

Example usage:

```
echo 'x 10 20 yay' | tr ' ' '\t'  >k.bed
odgi inject -i k.og -b k.bed -o - | odgi view -i - -g | vg view -dp - | dot -T png -o k_inj.png
```

[k.gfa.txt](https://github.com/pangenome/odgi/files/8597204/k.gfa.txt)

image

The injection functionality is pegged to a unit test that tests injection on paths going in both orientations over a small graph.


ekg:
Notes on setting up input to gggenes.

```bash
( echo 'x 10 20 A'; echo 'y 10 20 B'; echo 'y 25 35 C' ) | tr ' ' '\t'  >k.bed
odgi inject -i k.og -b k.bed -o - | odgi untangle -e 1 -R <(cut -f 4 k.bed ) -i -  \
    | tail -n+2 | awk 'BEGIN { print "molecule", "gene", "start", "end" } { print $1, $4, $2, $3 }' | tr ' ' '\t'  >x.tsv
```

Now in R

```R
require(tidyverse)
require(gggenes)
k <- read.delim('x.tsv')
ggplot(k, aes(xmin=start, xmax=end, y=molecule, fill=gene)) + geom_gene_arrow()
```

x

We see all the untangle targets. This is an extreme case just for testing---note the use of `odgi untangle -e 1`--- this makes a segment for every 1bp. In practice, we don't want to do that. It may be necessary to update `odgi untangle` to ensure that we always cut segments at target starts and ends.

ekg:
I corrected some problems with odgi untangle.

x

The last segment in a given path was dropped.

Target segments were mapping to other target segments when the jaccard was equal to 1. Technically not incorrect, but confusing. The self segment is now preferred.

docs/rst/commands/odgi_flip.rst
```diff
@@ -0,0 +1,65 @@
...
+
+SYNOPSIS
+========
+
+**odgi flip** [**-i, --input**\ =\ *FILE*] [*OPTION*]…
+
+DESCRIPTION
+===========
+
+The odgi flip command converts BED records against graph paths into new paths labeled by the BED record name.
```
subwaystation, line 17: This is about BED records, but I thought the tool just flips all paths? I am confused here.

PR #452, opened 2022-10-05
non-binary halotype matrix

ekg:
Useful for genotyping segdups / VNTRs.


AndreaGuarracino:
@ekg, why hasn't it been integrated into the master yet? Maybe, we could keep both ways (binary and non-binary) and put a flag to decide which output to emit?

Issue #440, opened 2022-08-09
Position command doens't output all paths


ASLeonard:
This is all done using the latest docker image.

I'm trying to lift coordinates through a pggb graph from one assembly to another with `odgi position`. This works fine for a given path (where regions.bed contains coordinates for ASM_1)

```
odgi position -i pggb.gfa -b regions.bed -r ASM_2
```

However, the documentation leads me to believe using `-R ref_paths.txt` containing ASM_2, ASM_3, ASM_4, etc. would then give output for every bed position for every ref_path in the `-R` file. However, it still only takes the first entry from the ref_paths. Similarly in the source code for odgi position, it seems to suggest not providing any ref_path name or file should default to using all paths in the graph https://github.com/pangenome/odgi/blob/bd54ab80ea03d864c82bd05599bf1dfa1777a72f/src/subcommand/position_main.cpp#L137-L139

Am I misunderstanding the approach here? Currently the working solution is just to invoke N calls of odgi position to liftover the entire bed file for the N other assemblies, but that seems inefficient.

Thanks,
Alex

AndreaGuarracino:
Hi @ASLeonard, would you please provide a (little if possible) reproducible example of your problem?

ASLeonard:
Even a subsetted og file was quite large, so if you need it I can provide it elsewhere. But the steps are
```
odgi paths -i pggb.smooth.og -L -l
#UCD	1	51098607
#Angus	1	51138358
#Highland	1	51174331
#OBV	1	52178328
#Brahman	1	51254507


cat breeds.txt
#UCD
#Angus
#Highland
#OBV
#Brahman


cat regions.bed
#UCD     1000   10000
#UCD     200000  3000000


odgi position -i pggb.smooth.og -R breeds.txt -b regions.bed
#UCD	1000	10000	UCD,1000,+     UCD,10000,+	+
#UCD	200000	3000000	UCD,200000,+	UCD,3000000,+	+
```
Where I would expect the additional lines
```
...
#UCD    200000	3000000	Highland,422079,+	Highland,3211922,+	+ 
...
```
which I can get by manually running 
```
for breed in UCD Angus Highland...
do
  odgi position -i pggb.smooth.og -r $breed -b regions.bed
done
```

ASLeonard:
Was looking a bit more into this but didn't make much progress. But here is some more information

[...]

PR #453, opened 2022-10-10
Fix position to print all paths for all bed regions

ASLeonard:
Fixes a presumed error reported in #440 where coordinates in a bed file are not translated into every path. Now given P paths in a graph and N lines of bed file coordinates, there is P * N lines of output. 

This is a fairly ugly fix by iterating `get_position` with an additional parameter to pick out each path, but is the best I can do given my understanding of handles. The change to `get_position` should not affect other calls to it, and I've tested it with both the `-r <path>` (unchanged behaviour) and `-R <paths_file>` (now fixed behaviour). I haven't looked at the other forms of input or liftover, so this is probably incomplete.

For a ~800 megabyte og file with 12 paths, I observed a 5x speed up (100 seconds old, 20 seconds new) with this fix versus having to call `odgi position -r <path>` in a for loop with a 5 line bed file. I assume loading the graph is takes a decent amount of time, so it is a big advantage to have this work internally.


AndreaGuarracino:
Thanks @ASLeonard, and sorry for the long delay. This is now addressed by #659 (merged), which adds an opt-in `--all-ref-positions` flag: with `-R` it lifts each BED range into every reference path, emitting one line per reference (P×N output), matching what you implemented here. Kept it opt-in so the default `-R` behavior (nearest ref) is unchanged. Closing as superseded — please reopen if #659 doesn't cover your case.

PR #458, opened 2022-11-22
Increment progress-meter of odgi sort less frequent

nsmlzl:
Improves the performance when using the progress indicator (`-P` flag) of odgi sort. Previously the progress-meter was incremented by every thread in each step operation which created a bottleneck.


AndreaGuarracino:
I get a nice ~20% boost with 12 threads with a few tests, but regardless of the detailed evidence, updating the log a little less will do just fine. Hot.

PR #462, opened 2022-11-28
Fix segfault when no progress flag set

nsmlzl:
#459 caused (on my machine) segmentation faults when progress indicator not used. This PR fixes this: Only increment progress meter when progress actually used. @AndreaGuarracino


AndreaGuarracino:
@nsmlzl thanks! It was my copy-paste mistake

PR #470, opened 2023-02-01
[odgi tension] Set the distance to 0 after each step instead of each path

tonyjie:
[odgi tension] Set path_layout_dist and path_nuc_dist to 0 for every step. Otherwise it will accumulate among the path, which will lead to little difference of tension for the later steps in the path.


subwaystation:
LGTM. Thanks for spotting this @tonyjie !

PR #472, opened 2023-02-03
script to generate release binaries

subwaystation:
This is a step forward to create binaries targeting different CPU architectures. A `bash` script is orchestrating the invocation of the appropriate binary.

- [x] Decide which AVX512 features to support by default. `-mavx512f` should be supported by all AVX512 architectures.
- [x] `-t/--threads` argument.
- [x] `-c/--clean` argument.
- [x] `-h/--help` argument.
- [x] Print which SIMD is actually used.
- [x] Invoke the script from a different folder.
- [x] Forward all arguments.


scripts/generate_release.sh
```diff
@@ -0,0 +1,149 @@
...
+  cmake -H. -Bavx2 -DEXTRA_FLAGS="-Ofast -pipe -mavx2" && cmake --build avx2 -- -j "$threads"
+  mv bin/odgi avx2
+  rm -r bin
+  mv avx2/odgi release/odgi_avx2
+fi
+
+if [[ ! -z "$AVX512" ]];
+then
+  echo "AVX512";
+  # TODO maybe only enable -mavx512f so it has the best compatibility - https://en.wikipedia.org/wiki/AVX-512#SIMD_modes - https://gcc.gnu.org/onlinedocs/gcc-6.1.0/gcc/x86-Options.html
```
subwaystation, line 106: Adding a single comment for Santiago.

subwaystation:
Evaluations are still in progress, but it makes sense to merge this anyhow.

Issue #475, opened 2023-02-10
odgi inject with overlapping bed entries


henrivkgt:
Hello,

Recently I have run into an issue with odgi inject when trying to inject genes and exons as paths in a graph. When running odgi inject using a bed file with overlapping coordinates (for instance when adding two transcript variants of the same gene), only the second entry is added correctly. The first entry can still be found in the graph when running odgi paths, but it is given a length of 0, and using odgi extract on the corresponding region does not extract the path. Is this intended? My current workaround is to separate the bed file into multiple files with only non-overlapping entries, then running odgi inject on all of them adding the paths recursively to the graph, but it is a bit inefficient as the graph has to be completely written to disk every time. 

Best,
Henri

AndreaGuarracino:
I can reproduce the issue. To me, it is triggered when two intervals share the same end coordinates:

```
echo -e "x\t1\t2\tA\nx\t1\t2\tB"
x	1	2	A
x	1	2	B
odgi inject -i test/t.gfa -b <(echo -e "x\t1\t2\tA\nx\t1\t2\tB") -o - | odgi view -i - -g | tail -n 3
P	x	1+,2+,3+,5+,7+,8+,10+,11+,13+,14+,16+,17+	*
P	A		*
P	B	2+	*

echo -e "x\t1\t2\tA\nx\t1\t2\tB"
x	1	2	A
x	1	3	B
odgi inject -i test/t.gfa -b <(echo -e "x\t1\t2\tA\nx\t1\t3\tB") -o - | odgi view -i - -g | tail -n 3
P	x	1+,2+,3+,4+,6+,8+,9+,11+,12+,14+,15+,17+,18+	*
P	A	2+	*
P	B	2+,3+	*
```

The issue is that paths are created, but no steps are added to them.

henrivkgt:
I checked with my graph and indeed it only happens when two bed entries share the same end coordinate.

AndreaGuarracino:
Thank you for confirming that!

Still investigating, and things are going worse.

```
cat t.gfa
H	VN:Z:1.0
S	1	ATCG
P	x	1+   *

odgi inject -i test/t.gfa -b <(echo -e "x\t1\t2\tA\nx\t1\t3\tB") -o - | odgi view -i - -g
H       VN:Z:1.0
S       1       A
L       1       +       2       +       0M
S       2       T
L       2       +       3       +       0M
S       3       C
L       3       +       4       +       0M
S       4       G
[...]

PR #476, opened 2023-02-14
`odgi inject`: use a key that is really unique across annotations

AndreaGuarracino:
This was leading to invalid paths when there were overlapping annotations to inject. See #475.


AndreaGuarracino:
Still buggy

```
cat t.gfa
H	VN:Z:1.0
S	1	ATCG
P	x	1+   *

odgi inject -i t.gfa -b <(echo -e "x\t1\t3\tA\nx\t1\t2\tB") -o - > /dev/null
[odgi::algorithms::inject_ranges] injection end point for interval B is not at a node boundary: off by 2 vs 3 on a node 1bp long
```

AndreaGuarracino:
Inverting the end coordinates works:

```
odgi inject -i t.gfa -b <(echo -e "x\t1\t2\tA\nx\t1\t3\tB") -o - > /dev/null
```

PR #485, opened 2023-03-16
`odgi flip`: avoid introducing useless edges and losing edges from the original graph

AndreaGuarracino:
This PR introduces a different way to construct the flipped graph:
- copy all edges from the input graph
- flip paths
- check missing edges (due to flipped paths)
- add missing edges (if any)

In the first tests, this avoids introducing edges that are not used in the flipped paths and avoids losing edges that are present in the input graph.

### DIFF
--- a/src/algorithms/flip.cpp
+++ b/src/algorithms/flip.cpp
@@ -9,25 +9,24 @@ using namespace handlegraph;
 void flip_paths(graph_t& graph,
                 graph_t& into,
                 const std::vector<path_handle_t>& no_flips) {
-    // for each path, find its average orientation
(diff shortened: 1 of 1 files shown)


anshumanmohan:
Thanks for this, @AndreaGuarracino! I can recreate the fact that it now (correctly) does nothing for the graph `LPA.gfa`. However, it now does the same fishy thing for `note5.gfa`.

When I run flip on note5, the only meaningful change is that
`L	3	-	4	+	0M`
becomes
`L	4	-	3	+	0M`.
The paths, segments, and other links remain the same.

The correct behavior (if I understand flip correctly) is to do nothing.

PR #474, opened 2023-02-06
prevent quadratic untangle cut point explosions

ekg:
That's the premise of this PR. It seems that untangling isn't behaving as expected in deep loops, and this branch will have hacks required to figure out why. It may be due to a quadratic expansion of the number of possible cut points.

The current patch makes a change that does affect the untangle cut points. We need to validate that the output makes sense.

PR #486, opened 2023-03-16
Untangly cleanup v2

ekg:
Working on this. Idea is to push more of the core untangle segmentation into vectors, trading time for space.


AndreaGuarracino:
Closing to consolidate the untangle work: this overlaps #474 (prevent quadratic untangle cut point explosions), which is the more recently updated branch. Rather than maintain two parallel untangle refactors from the same era, we'll carry the work forward in #474. The vector-based segmentation idea here can be folded in there. Reopen if you'd prefer to keep this one as the base instead.

PR #494, opened 2023-04-12
Portability fixes to build with Clang, on macOS, and on ARM

sampsyo:
This may have been misguided, but I was interested in building & running odgi on my (macOS, AArch64) laptop. My previous attempt (documented in #350) was to build with Clang on Linux; I found a bunch of GCC-isms then. I found some more this time around and also found a few Linux-isms. Fortunately, none of them were that hard to fix, so I plowed through and repaired them all.

I left more detailed commentary in the individual commit messages, but here are the highlights:

* Clang detected several integer type disagreements in uses of `std::min` and `std::max`. These all essentially had to do with using the length of a `std::string` or `std::vector` on one side and `uint64_t` on the other. The fix was to either fix a variable declaration to be `size_t` or do a conversion.
* Some header on my system declares a type called `off_t`. It wasn't hard to avoid a naming conflict with some cleverness (I didn't have to rename odgi's `off_t`).
* One file had a manual check that the target was x86_64. I've turned this into a warning, which might be a problem someday.
* `get_current_dir_name` is a GNU-ism; I've replaced it with the standard `getcwd`.

FWIW, making this build on macOS in the end is pretty easy. We do have to use Homebrew's LLVM to get OpenMP working. You just need to `brew install llvm jemalloc` and then:

```
$ CC=/opt/homebrew/opt/llvm/bin/clang CXX=/opt/homebrew/opt/llvm/bin/clang++ LDFLAGS=-L/opt/homebrew/lib cmake . -Bbuild
$ CC=/opt/homebrew/opt/llvm/bin/clang CXX=/opt/homebrew/opt/llvm/bin/clang++ LDFLAGS=-L/opt/homebrew/lib cmake --build build -- -j
```

Just for fun, building all of odgi from scratch on my moderately loaded machine (MBP w/ M1 Max and 32 GB RAM) took 1m6s. Running `odgi build` on the 4.2 GB `chr8.pan.gfa` takes 2m4s. Those times are in the same ballpark that I recall from our big fancy Intel server we have in the lab.

It's possible I've done some wrong things here, so feedback would be very welcome!


sampsyo:
Yeah; sounds great!

martin-g:
Hi,

I tried to add ARM64 builds for the Bioconda package but it fails with a segfault in one of the tests - https://github.com/bioconda/bioconda-recipes/pull/48682#issuecomment-2415887709
I'd be glad to test any suggestions!

Issue #426, opened 2022-06-24
Make odgi compile on Mac


George-du:
Dear odgi team,
      I want to compile odgi on my Mac, can you suggest something about installing odgi on Mac arm64.
Thanks,
    Du

AndreaGuarracino:
Hi @George-du, neither of us owns a Mac, but we would be happy to work with you in trying to make `odgi` compilable there if you can. Which is the current starting point (problem)?

AndreaGuarracino:
@George-du, have you tried something? It would be interesting having details about it, also regarding failures!

George-du:
Dear odgi team,
  I have compiled it in a linux environment. I will try to compile odgi  on Mac arm64 soon.
  Thanks,
Du

anshumanmohan:
@George-du, you may enjoy https://github.com/pangenome/odgi/pull/494!

kojix2:
Hello.
I followed the instructions #494 above and compiled odgi on my M2 Mac mini. 

```sh
brew install llvm jemalloc
```

```sh
CC=/opt/homebrew/opt/llvm/bin/clang CXX=/opt/homebrew/opt/llvm/bin/clang++ LDFLAGS=-L/opt/homebrew/lib cmake . -Bbuild
CC=/opt/homebrew/opt/llvm/bin/clang CXX=/opt/homebrew/opt/llvm/bin/clang++ LDFLAGS=-L/opt/homebrew/lib cmake --build build -- -j
```

I had to add "#include <sstream>" to `odgi/src/algorithms/progress.hpp` because I got the following error.

```
odgi/src/algorithms/progress.hpp:68:31: error: implicit instantiation of undefined template 'std::basic_stringstream<char>'
            std::stringstream buffer;
                              ^
```

PR #512, opened 2023-06-20
Add missing library

kojix2:
Hi odgi team!
This pull request is a change I needed to edit when I compiled odgi on my Mac mini. 
I know almost nothing about C++ so this change may not be appropriate. If so, please feel free to close it.
Thank you.

See #426


AndreaGuarracino:
I don't see anything dangerous, thank you for the fix!

PR #538, opened 2023-11-18
depth: option to use unique window depths

kdm9:
Hi folks,

I needed an equivalent to `odgi depth --windows-in` that considered only unique depth, so I've implemented this behind a CLI flag (`--window-unique-depth`).

best,
Kevin


subwaystation:
Thanks for this @kdm9 :)

The PR looks good to me.

PR #539, opened 2023-11-29
make svg drawing work with colors

ekg:
Allows us to add BED annotations to SVGs, for pretty plots like this:

image

The scaling of everything in SVG is a bit insane. It could use some work.


ekg:
This also allows us to modulate the width of the rendered nodes.

subwaystation:
:hot:

PR #547, opened 2024-01-16
write highlights over the rasterized image

ekg:
This appears to work, but it's a bit weird. We collect non-gray segments and then rasterize them on top of the existing image.


ekg:
The problem is that we were getting renderings like this:

image

After this change, we get this:

image

PR #558, opened 2024-02-10
`odgi draw`: add node sparsification factor for SVG output

AndreaGuarracino:
This allows us to make smaller SVG output for big graphs

**-f 0.1**
x_0 1

**-f 0.3**
x_0 3

**-f 0.5**
x_0 5

**-f 0.7**
x_0 7

**-f 0.9**
x_0 9

**-f 1.0** (no node sparsification)
x_1 0

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ docs/sphinx_build_man
 docs/_build
 Testing/
 .idea/
+.vscode/
+.cmake/
+cmake-build-debug/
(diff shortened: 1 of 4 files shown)


ekg:
A related trick would be to make the nodes proportionally longer, filling in the gaps.

AndreaGuarracino:
Machiavellian. Not so clean with strong sparsification (layouts with stretched nodes to the left):

*-f 0.1*
image

*-f 0.3*
image

*-f 0.5*
image

*-f 0.7*
image

*-f 0.9*
image

ekg:
I wonder if the sparsification could be driven to remove short nodes first. That might retain the structure while greatly reducing the number of nodes.

PR #566, opened 2024-03-18
Upgrade pybind11 to current head

JervenBolleman:
The pybind11 submodule is very old and not up to date with some of the current cmake rules.
Use the current pybind11 head, and avoid errors around building.


subwaystation:
LGTM!
Thanks @JervenBolleman.

PR #585, opened 2024-07-31
`odgi flip -r/--ref-flips`: flip with respect to a set of paths (used as references)

AndreaGuarracino:
This helps preserve coordinate systems on paths (ref paths) that should not be reversed and orient all other paths on those ref paths.

Input graph:
```
odgi viz -i x.og -o x.png -z
```
x

Flipped graph:
```
odgi flip -i x.og -o - | odgi viz -i - -o x.png -z
```
x

Flipped graph by not flipping the `chm13#chr6:31825263-31908622` path:
```shell
odgi flip -i x.og -o - -n <(echo chm13#chr6:31825263-31908622) | odgi viz -i - -o x.png -z
```
x

Flipped graph w.r.t. the `chm13#chr6:31825263-31908622` path:
```shell
odgi flip -i x.og -o - -r <(echo chm13#chr6:31825263-31908622) | odgi viz -i - -o x.png -z
```
x

### DIFF
--- a/src/algorithms/flip.cpp
+++ b/src/algorithms/flip.cpp
@@ -8,10 +8,13 @@ using namespace handlegraph;
 
 void flip_paths(graph_t& graph,
                 graph_t& into,
-                const std::vector<path_handle_t>& no_flips) {
+                const std::vector<path_handle_t>& no_flips,
+                const std::vector<path_handle_t>& ref_flips) {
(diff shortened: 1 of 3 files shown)


AndreaGuarracino:
It may also be useful to flip the graph nodes, not only paths.

PR #599, opened 2024-12-05
WIP: Improving random number generation on gpu

Shihab-Shahriar:
Hi,

This PR attempts to improve the random number generation component of GPU accelerated `odgi-layout`. It replaces the current generator with a counter-based one: Philox. These can be created, used and discarded all from within thread registers, completely eliminating any global memory bookkeeping, communication etc. It also makes the code reproducible and it is statistically very robust.  

On a V100 GPU, I noticed performance improvement of around 17%. On A100-SXM4 with significantly higher memory speed, the effect was a little less pronouned. But it was still around 10%.

This is a work-in-progress, it needs a bit of cleanup. I am more than happy to hear you suggestions, feedback and incorporate them back into the code. 

On a relevant note, I am still trying to figure things out, but it seems the current code assumes there can be at most one thread block running on a SM. There appears to be one `curandStateXORWOWCoalesced_t` object _per SM_, so if there is more than one block on a SM, that could create race condition. 

Thanks to @tonyjie, @subwaystation for helping me set this odgi up.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -637,6 +637,7 @@ add_library(odgi_objs OBJECT
 )
 if (USE_GPU)
   target_sources(odgi_objs PRIVATE "${CMAKE_SOURCE_DIR}/src/cuda/layout.cu")
+  # target_compile_options(odgi_objs PRIVATE $<$<COMPILE_LANGUAGE:CUDA>:-Xptxas="-v">)
 endif (USE_GPU)
 
 set(odgi_DEPS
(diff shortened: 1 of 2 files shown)


tonyjie:
Could you first check the layout figure generated by our current `odgi-layout` and yours "visually" to make sure they are similar? This is done by `odgi draw`, check [here](https://github.com/tonyjie/gpu_pangenome_layout_artifact/blob/a0e2c2166015e5adfa56f81667fd38e83f21e5f7/experiments/evaluation/run_gpu_layout_all.sh#L39) for reference. 

Later you might also check the quantitative metric of the layouts generated by your PRNG. Check this [`odgi tension`](https://github.com/tonyjie/gpu_pangenome_layout_artifact/blob/a0e2c2166015e5adfa56f81667fd38e83f21e5f7/experiments/path_stress/run_path_stress_verify.sh#L27) command to evaluate the metric. But note that this is not merged into the pangenome/odgi yet. It is available in [this branch](https://github.com/tonyjie/odgi/blob/jiajie/pr_gpu_layout/src/subcommand/tension_main.cpp)

tonyjie:
And is it possible to highlight the key differences/advantages between your PRNG and cuRAND? I took a look at the paper, but not sure if I totally get it. Thanks!

Shihab-Shahriar:
To answer your other question, OpenRAND is a library that implements several well-known PRNGs. For example, Philox used here was introduced in 2014 in [this paper](https://dl.acm.org/doi/abs/10.1145/2063384.2063405?casa_token=3XjSswU-WswAAAAA:MBXAOSeXr_eMx23piKm5eqRPE0rwnyW251Rz9KcFfOhj_Q7sdA3SmXYmrHOiMKyt1D2E2O6fdv3y). These cryptography-inspired generators tend to be more statistically robust and GPU-friendly than alternatives like Xorwow.

As a library, OpenRAND API does provide a key performance benefit over cuRAND- in that it doesn't require a separate random state initialization kernel and memory load/store when they are subsequently used. These steps are completely unnecessary for a generator like Philox. That can save quite a bit of memory communication.

PR #626, opened 2026-02-10
fixed the extract_selected_loci.rst documentation

mirelaminkova:
Hello, 

Recently, I followed the extract selected loci tutorial from odgi. However, with the given settings, users cannot reproduce the same output. Here are some suggestions + a fix on the odgi viz command.


AndreaGuarracino:
Thank you for taking care of the docs!

PR #631, opened 2026-04-30
Fix compilation on aarch64

ankane:
Makes `__target__("arch=x86-64")` conditional and removes `-msse4.2` (as `-march=native` will automatically apply it for platforms that support it).


AndreaGuarracino:
Thanks @ankane!

ankane:
Thank you!

